Issue #7, opened 2022-02-03
Change name of package (formattable - with 2 t's is an existing CRAN package)


gmbecker:
for discussion with @waddella

kpagacz:
I assume this can be closed?

pawelru:
Nope - still in progress. Package name still needs to be changed with everything that follows

PR #14, opened 2022-03-24
rename package to formatters

pawelru:
close https://github.com/insightsengineering/formatters/issues/7


junlueZH (approved):
no instance of `formatable` can be found.

Is the branch name problematic without the `@main` suffix?

pawelru:
> Is the branch name problematic without the `@main` suffix?

I don't think so. `main` is the default branch that staged-deps would use

PR #24, opened 2022-05-10
update description license

pawelru:


gmbecker:
We previously had a LICENSE file but CRAN specifically asked us to remove it, because we are using a standard license, during the initial submission process. Roche is listed as the copyright holder in the DESCRIPTION file.

Closing this.

PR #46, opened 2022-09-22
Wrapping titles and footnotes automatically

Melkiades:
The points to discuss are wether the current implementation:

- [x] fits the desired needs (@shajoezhu).
- [x] needs to propagate or have more parameters (please notice that to see the effects of the implementation it is needed to use `toString` as it is done in the newly designed tests).
- [x] works fine also in exports (such as pdfs) as this is a working solution for direct prints to the terminal only atm (this cannot be solved yet as it depend on the parameter propagation in `rtables` version of `toString` and where it is used)
- [x] should deal with relevant parameters (`nchar_max`) to do so.
- [x] is safe for extreme cases (one is already considered), and the various input checks (do we need `checkmate`?) -> Opened an issue for checking inputs and considered extreme cases where words are extremely large in the following comments.


R/tostring.R
```diff
@@ -154,6 +186,32 @@ setMethod("toString", "MatrixPrintForm", function(x,
...
+      )
+      stop(err_msg)
+    }
+    if (all(to_keep)) break
+    gen_tmp_list[[length(gen_tmp_list)]] <- tmp_split[to_keep]
+    gen_tmp_list[[length(gen_tmp_list) + 1]] <- tmp_split[!to_keep]
+  }
+  paste0(sapply(gen_tmp_list, function(y) {
+    paste0(y, collapse = " ")
+  }), collapse = paste0(c("\n", rep(" ", txt_indent)), collapse = ""))
```
gmbecker, line 211: IF we are going to do this, we already have utilities to indent and pad strings, I believe, please use these (again, if we implement this functionality ourselves at all, which i hope not to)

gmbecker (commented):
Whether we want to indent after wrapping or not is something of an open question, I'd say focus on the wrapping aspect of it for now and we can add the indent if needed once its in.

R/tostring.R
```diff
@@ -154,6 +186,32 @@ setMethod("toString", "MatrixPrintForm", function(x,
 ##     x
 ## }
 
+splitter_new_line <- function(x, nchar_max = 80, txt_indent = 1) {
```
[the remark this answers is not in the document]
Melkiades: My apologies! I did not know there was a way to do it in `base`. I checked all the options and it seems what we need (also indent as I implemented it is available under the `exdent` variable (which is what I think is most needed for titles and footers). I want to reference here issue [rtables#373](https://github.com/Roche/rtables/issues/373), so to connect the two things for the future

R/tostring.R
```diff
@@ -48,13 +48,17 @@ default_hsep <- d_hsep_factory()
 #' of the table-like object reprseented by `x`
 setMethod("toString", "MatrixPrintForm", function(x,
                                                   widths = NULL,
+                                                  indent_titles = FALSE,
```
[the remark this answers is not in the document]
Melkiades: It makes perfect sense. I will add a `wrap` logical and a `max_width` numeric. What do you think should be the default for the second one? and what should be done in case some parts of the text have "words" that are longer than the allocated horizontal space? In my implementation, I was throwing an error and the two things above are related because also my test cases produce only a small table, and if you use `ncchar` as default, the horizontal space allowed is almost always smaller than the added title. I think the best would be to use 80 (terminal std) as default. This will change a lot of tests downstream though (maybe not too many as it is only title and footer)

R/tostring.R
```diff
@@ -154,6 +173,19 @@ setMethod("toString", "MatrixPrintForm", function(x,
 ##     x
 ## }
 
+wrap_list <- function(txt_lst, max_width) {
+  if(all(!sapply(txt_lst, grepl, pattern = "\n")) &&
```
Melkiades, line 177: I think should output a warning, because the user asked to wrap but inserted already some `\n`

R/tostring.R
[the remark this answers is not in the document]
gmbecker: The default should either be nothing, or the current value of `options("width")` in interactive print to the console settings. In any pagination aware context it should instead be the value of `cpp`, if that is set.

You're right that this will break a lot of tests. This doesnt mean we shouldnt do it, but it does mean we need to be careful of when we put this change in. Its probably far enough from a release sprint that we can get away with it, but if not, it would need to wait and be added to to the next (in person) PI planning board. We can develop the feature in a branch in the meantime though so it doesn't necessarily mean we can't work on this.

Also, we can do this in two steps. Put the machinery in but leave the default as is (ie nothing), and then later when we're ready to flip the switch, we simply change the default.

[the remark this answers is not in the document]
gmbecker: As for "words" that take up more than the allotted space, that is indeed an issue we'll need to think about, as `base:strwrap` unfortunately doesn't help us there.  The case I forsee where that may happen is file paths in the provenance footer (it should be exceedingly rare or even impossible with real words in any realistic setting):

```
> strwrap("/my/path/to/something_really/long/that/is/too/long", width = 10)
[1] "/my/path/to/something_really/long/that/is/too/long"
```

R/tostring.R
gmbecker: Once wrapping is in, I don't think users should be putting `\n` in there title/footer information, agreed.

The only one that can't take a vector is main title. Everything else already takes a vector so if they want manual break points they can do that.

R/tostring.R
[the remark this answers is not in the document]
Melkiades: I think we could start dealing with the above cases (that, as you said, certainly will appear) as soon as they do appear. For the moment, I think we can keep it as it is.

R/tostring.R
Melkiades: I see, that is why I did not need to fix a `\n` in when unlisting the output. as it just outputs a vector following the newlines. In other words, it is not important to preserve the structure of `main_title`, `subtitle`, and so on. Maybe in the future, if users desire to have spacing between those, we could modify it accordingly.

Melkiades:
@gmbecker I could proceed in propagating this in `tt_toString.R` if you agree to the status of things. I also added some documentation to the function with some details about the idea of newline character (I was not aware you could do that tbh)

gmbecker:
> @gmbecker I could proceed in propagating this in `tt_toString.R` if you agree to the status of things. I also added some documentation to the function with some details about the idea of newline character (I was not aware you could do that tbh). From this we need to:
> 
> * [ ]  propagate two vars (`wrap` and `max_width`)

I don't think we can call it `wrap` because it doesn't control word-wrapping for cell values (its correct that it doesn't, as we talked about, I just thinkt he name needs to change. `tf_wrap` maybe? Lets discuss it tomorrow.

> * [ ]  decide how to proceed with indentation
> * [ ]  Should I include also `ref_footnotes` in the current wrapping? is it for references?
Yes they should wrap to the same width as the rest of the title/footer materials

> * [ ]  create a new issue to discuss possible workarounds for long paths or long "words" that could exceed the inserted width.

I'd say we just do it. Do a check after attempted wrapping and if it didn't wrap, and is still too long, just do a blind hard break at the width.

Melkiades:
I encountered a weird case to me: 
```r
a <- "sub titl e is"
strwrap(a, 4)
[1] "sub"  "titl" "e"    "is"  
```
while I would expect `"sub"  "titl" "e is"`. Is this the right behavior?

DESCRIPTION
```diff
@@ -9,14 +9,17 @@ Authors@R: c(
...
 Depends: methods, R (>= 2.10)
-Imports: htmltools, grid
+Imports: 
+  grid,
+  htmltools,
+  stringi
 Suggests:
   dplyr,
-  testthat,
+  gt (>= 0.7.0),
```
Melkiades, line 18: this should be alphabetical

gmbecker:
> ```r
> sub titl e is"
> ```



> I encountered a weird case to me:
> 
> ```r
> a <- "sub titl e is"
> strwrap(a, 4)
> [1] "sub"  "titl" "e"    "is"  
> ```

Right, so this is related to a note in the docs:

>  ## Note that messages are wrapped AT the target column indicated by
>    ## 'width' (and not beyond it).
>   ## From an R-devel posting by J. Hosking <jh910@juno.com>.
>     x <- paste(sapply(sample(10, 100, replace = TRUE),
>                function(x) substring("aaaaaaaaaa", 1, x)), collapse = " ")
>     sapply(10:40,
>           function(m)
>            c(target = m, actual = max(nchar(strwrap(x, m)))))
>

```
> ?strwrap
> x <- paste(sapply(sample(10, 100, replace = TRUE),
+                 function(x) substring("aaaaaaaaaa", 1, x)), collapse = " ")
>      sapply(10:40,
+             function(m)
+             c(target = m, actual = max(nchar(strwrap(x, m))))
+ )
       [,1] [,2] [,3] [,4] [,5] [,6] [,7] [,8] [,9] [,10] [,11] [,12] [,13]
target   10   11   12   13   14   15   16   17   18    19    20    21    22
actual   10   10   11   12   13   14   15   16   17    18    19    20    21
       [,14] [,15] [,16] [,17] [,18] [,19] [,20] [,21] [,22] [,23] [,24] [,25]
target    23    24    25    26    27    28    29    30    31    32    33    34
actual    22    23    24    25    26    27    28    29    30    31    32    33
       [,26] [,27] [,28] [,29] [,30] [,31]
target    35    36    37    38    39    40
actual    34    35    36    37    38    39
> ?strwrap
> x
[1] "aaaa aaaaaaaaa aaaaaaaaa aaaa aaaaaaaaa aaaaaaaa aaaaaaaaaa aaaaaa a aaaaaa aaaaaaaa aaaaa aaaaaaaaa a aa aaaaa aaaa aaaa aaaaaa aaa aaaaaaa aaaaaaa aaaaa aaaaaaa aaaa aaaaa aaaaaaaaa aaaaaaaaaa aaaa aaaaaaaaa aaaaaaa aaaaaaaaa aaaaa aaaaaaaaaa aaaaaa a aaaaa aaaaaaaaaa aaaaaa a aaaaaaa aaaaaaaa a aaaaaa aaaaaaaaa aaa aaaaa aa aaaaaaa aaaaa aaaaaaa aaaaaaa aaaaaaa aaaaa aaaa aaaaa aaaaaa aaaaaa aaaaaaaaa aaaaaaaaa aaaa aaaaaaaaaa aaaaaaaaaa a aa aa aa aaaaaaaa aaaaa aaaaa aaaaaaaaaa aaaaaaaaa aaa aaaaaaaaaa aaaaaaaaa aa aa a aaaaaaa aa aaaaaaaaa aaaaa aa aaaaaaaaaa aaa aaaa aaaaaaaa aaaaaaa aaa aaaaaaaaa aa aaaaaaaaa aaaaaaa aaaaaaaaaa aaaaaa aaaa aaaaaa aaaaaaaa aaaa aaaaaaaaaa"
```

So basically what that means, from what I can tell, is that the `n`th character is going to bet he first *after* the wrapping, rather than the last before it, where `n` is the specified width. The reason `titl` is still there is because strwrap leaves intacts words that are "too big" as we found before.  Given that we know this about `base::strwrap` we should be able to work around it with no problem. 

Basically we need to do our manual breaks as we're doing now, and then specify width as 1 larger for the purposes of `base::strwrap` and everything should work out.


> while I would expect `"sub" "titl" "e is"`. Is this the right behavior? Also, this makes a bit useless my last fix that aimed at getting rid of the lonely `e` in this case. I actually found an alternative to this that solves the problem as needed:
> 
> ```r
> stringi::stri_wrap(a, 4)
> [1] "sub"  "titl" "e is"  
> ```
> 
> Let me know if it is fine for me to add this to the suggests @gmbecker

I'd rather we work around base::strwrap, rather than adding a strong dependency on `stringi` (it would not be able to be a `Suggests` moving forward, I don't think)

Melkiades:
Ok, I did not know about the workaround. It makes perfect sense now. `stringi` removed.

gmbecker:
@Melkiades please rebase and make any necessary changes for this PR to be compatible with the table_inset functionality I just added

R/tostring.R
```diff
@@ -36,29 +36,54 @@ default_hsep <- d_hsep_factory()
...
-#' @param widths (proposed) widths for the columns of \code{x}
+#' @param widths numeric (or NULL). (proposed) widths for the columns of
+#'  \code{x}. The expected length of this numeric vector can be retrieved with
+#'  `ncol() + 1` as the column of row names must also be considered.
 #' @param hsep character(1). Characters to repeat to create header/body
-#' separator line.
-#' @exportMethod toString
+#'  separator line.
+#' @param tf_wrap logical(1). Should the texts like title and footnotes be wrapped?
+#' @param max_width integer(1) (or NULL). If NULL, it will be assigned to
```
gmbecker, line 47: This description doesn't mention that it only applies to title and footer materials and not the body of the table.

NEWS.md
```diff
@@ -9,6 +9,7 @@
 ## formatters 0.3.3.3
  * length-zero values of `na_str` are now silently interpreted as the default value (`"NA"`),
    as are any missing values in `na_str` of length > 0.
+ * adding wrapping of titles, subtitles, and footers.
```
gmbecker, line 12: This should be under a new 0.3.3.6 heading, not under 0.3.3.3 (since it is being added after 0.3.3.5 hit main)

R/tostring.R
```diff
@@ -36,29 +36,54 @@ default_hsep <- d_hsep_factory()
...
+#'  `options("width")`. Otherwise, it is also possible to use `c("auto")` to
+#'  automatically use the table width.
+#'
+#' @details When wrapping texts such as title, subtitle, or footnotes, please
```
gmbecker, line 51: This text will need to be massaged but I can do that after merge.

DESCRIPTION
```diff
@@ -9,14 +9,16 @@ Authors@R: c(
 Description: We provide a framework for rendering complex tables to ASCII, and a set of formatters for transforming values or sets of values into ASCII-ready display strings.
 License: Apache License 2.0
 Depends: methods, R (>= 2.10)
-Imports: htmltools, grid
```
gmbecker, line 12: Version bump (devel component) for the new feature to 0.3.3.6 please

gmbecker (changes requested):
Pretty much there, couple of tweaks.

R/tostring.R
Melkiades: Yes of course it has to be completed into the bigger picture of the function

Melkiades:
I think we are there. I modified slightly the `.footer_inset_helper` so to use another variable instead of `x` so the wrapping preprocessing can still be done.

R/tostring.R
```diff
@@ -165,20 +209,22 @@ setMethod("toString", "MatrixPrintForm", function(x,
...
+    prvf <- footers_v$prov_footer
+    rfn <- footers_v$ref_footnotes
+    footer_txt <- .do_inset(rfn, inset)
     if(any(nzchar(footer_txt)))
         footer_txt <- .inset_div(footer_txt, div, inset)
-    if(any(nzchar(all_footers(x)))) {
-        if(any(nzchar(prov_footer(x))))
-            provtxt <- c(if(any(nzchar(main_footer(x)))) "",
-                         prov_footer(x))
+    if(any(nzchar(footers_v))) {
```
gmbecker, line 220: This is a list, so please don't call nzchar on it directly, the result can be counter-intiutive:
```
> nzchar(list(""))
[1] FALSE
> nzchar(list("", c("", "")))
[1] FALSE  TRUE
```

Other than that, having the inset happen on the wrapped things makes sense to me. Be sure you're wrapping to the correct widths though, titles should wrap as before, but main footer and referential footnotes have `inset` less space and need to be wrapped accordingly

gmbecker (changes requested):

gmbecker (approved):

PR #52, opened 2022-10-07
adding col_gap stress test for auto

### LINKED ISSUE #51: Adding more stringent tests for `"auto"` option
Reference https://github.com/Roche/rtables/issues/409

Melkiades:
fixes #51

### DIFF
--- a/tests/testthat/test-txt_wrap.R
+++ b/tests/testthat/test-txt_wrap.R
@@ -45,8 +45,8 @@ test_that("tf_wordwrap and table inset work (including together)", {
 
 test_that("toString() throws a warning when newline is in string", {
     bmf <- basic_matrix_form(iris)
-    bmf$main_title <- "some\nvery\nspacious\ntitle"
-    bmf$prov_footer <- "some\nvery\nspacious\nfooter"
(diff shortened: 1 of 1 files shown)


tests/testthat/test-txt_wrap.R
```diff
@@ -54,38 +54,72 @@ test_that("toString() throws a warning when newline is in string", {
 
 test_that("works with words that are too big (no warning)", {
     bmf <- basic_matrix_form(mtcars[1:2, 1:2])
-    bmf$col_gap <- 3
-    bmf$table_inset <- 1
-    bmf$main_title <- strrep("a", 30)
```
gmbecker: Please use the setters (e.g., `main_title(bmf) <- strrep("a", 30)`, etc) instead of direct assignment so that these tests are robust to us changing the MatrixPrintForm class in the future (which may or may not happen, but its a good idea not to penetrate abstraction when you don't need to for that reason).

Other than that this looks good

gmbecker (changes requested):

gmbecker (commented):
Looks good to me. Thanks

gmbecker (approved):
Looks good

PR #62, opened 2022-11-01
Update workflows

cicdguy:
Update package so that it conforms with all CI pipelines.


edelarua (approved):
Everything looks fine to me!

PR #74, opened 2022-11-30
Fix and extend various pagination support features.  closes #61

### LINKED ISSUE #61: Flexible column width (including column header) adjustment
Users can specify width for each column, and if they don’t, rlistings will get maximum content length from each column and use that as the column width
The width of all columns combined will be compared with the pre-specified or default pagesize, and if it exceeds the pagesize, some columns will be shown on subsequent pages (see Example 2 below)
Example 1: for a variable with label of “AA BBBB CCC”, it contains text of “abcdefg hijklmn opqrst”, user specifies a column width of 6, the column will be shown in the listing as:

AA BBB
B CCC
abcdef
g hijk
lmn op
qrst

If user doesn’t specify a column width, the width will be set at 22, which is the length of text “abcdefg hijklmn opqrst”
Example 2: 4 variables, each has a width of 40, pagesize set as 100, then the output will show first 2 variables on 1st page, another 2 variables on 2nd page
Additional note: A word can be broken into 2 lines (see Example 1 above for word “abcdefg”) and we currently do not consider preserving word when doing text wrappings as a must-have

gmbecker:
[skip vbump]


gmbecker:
@cicdguy Can you please look into why this was merged in. I turned on auto-merge but I was under the impression that that would only take effect once all the required checks were passed.

That doesn't appear to be the case (though the failure is in git, rather than in the package, which is also weird?)

Issue #77, opened 2022-12-12
table sections after last non-na section divider are not printed


gmbecker:
See https://github.com/Roche/rtables/issues/485#issuecomment-1344570085

PR #78, opened 2022-12-12
Fix missing table portions after last section div (#77) vbump

gmbecker:
[skip vbump]


gmbecker:
@cicdguy looks like more issues with auto-merge and skipped checks

PR #79, opened 2022-12-12
reintroduce obj_na_str after regression. vbump

gmbecker:
[skip vbump]


R/generics.R
```diff
@@ -285,6 +285,21 @@ setMethod("obj_format<-", "ANY", function(obj, value) {
   obj
 })
 
+#' @rdname lab_name
```
shajoezhu, line 288: Hi @gmbecker , it took me a while to guess what "lab_name" means, I guess it is 'label_name'? I was wondering if this what you meant? if it is, we can spin name another issue and replace it to ‘label_name'?

shajoezhu:
Hi @gmbecker , I was wondeirng if there is some background issue I could refering this to? Thanks

R/generics.R
gmbecker: this only controls the physical name of the Rd file within the directory structure, so it really doesn't matter what it is (and thus doesn't seem worth changing)

gmbecker:
@cicdguy the linter blocked merging because it makes changes but skips actions which means its impossible to meet the requirements

R/generics.R
```diff
@@ -498,7 +499,8 @@ setGeneric("table_inset<-", function(obj, value) standardGeneric("table_inset<-"
 
 #' @rdname table_inset
 #' @export
-setMethod("table_inset<-", "MatrixPrintForm",
+setMethod(
+  "table_inset<-", "MatrixPrintForm",
```
gmbecker, line 503: @cicdguy in paticular, this change that the linter is trying to force is actively bad and I do not want it.

R/generics.R
shajoezhu: hi Gabe, this is just a minor comment, definitely not blocking this PR. I was just thinking if there are ways that we could improve, and make the code easier to read. :)

shajoezhu (approved):
lgtm! thanks @gmbecker

Issue #84, opened 2023-01-13
Row label wrapping with identical indentation


Melkiades:
Ref: https://github.com/Roche/rtables/issues/469

PR #83, opened 2023-01-10
Pagination documentation, turn on md for roxygen to fix docs.

gmbecker:

PR #86, opened 2023-01-16
Fix indentation and row label wrapping

### LINKED ISSUE #85: Error when more than one rowname is to be wrapped with indentation
The following code:
 ```r
set.seed(1)
tiris <- iris[sample(seq_len(nrow(iris)), 100), ]
library(rtables)
a <- basic_table() %>%
  analyze("Petal.Width",
          function(x) list("what a long string" = rcell(1.2), "also very long long" = rcell(1.3)), indent_mod = 1) %>%
  build_table(tiris) %>%
  print()
```
produces:
```
                       all obs
——————————————————————————————
  what a long string     1.2  
  also very long long    1.3  
```
while the following:
```r
b <- matrix_form(a, indent_rownames = T)
cat(toString(b, widths = c(15, 5)))
```
crashes (without indentation this does not crash and it works):
```
Error in (function (x, n, just = c("center", "left", "right"))  : 
  "  also very long" has more than 15 characters
```

This is related to #84

Melkiades:
Fix #85 #84

### DIFF
--- a/R/tostring.R
+++ b/R/tostring.R
@@ -105,7 +105,7 @@ setMethod("toString", "MatrixPrintForm", function(x,
                                                   col_gap = x$col_gap,
                                                   hsep = default_hsep()) {
   mat <- matrix_form(x, TRUE)
-
+  browser()
   inset <- table_inset(mat)
 
   if (is.null(widths)) {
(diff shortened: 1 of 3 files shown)


R/tostring.R
```diff
@@ -105,9 +105,8 @@ setMethod("toString", "MatrixPrintForm", function(x,
                                                   col_gap = x$col_gap,
                                                   hsep = default_hsep()) {
   mat <- matrix_form(x, TRUE)
-  browser()
   inset <- table_inset(mat)
-
+  if(tf_wrap) browser()
```
gmbecker: Need this removed before we can merge.

gmbecker (changes requested):

R/tostring.R
Melkiades: I think I removed in commit [forgot browser](https://github.com/insightsengineering/formatters/pull/86/commits/a0e7bd8e4712caa59ea3fb6b8f968c005372d6b0)

but I do not understand the errors that are still blocking this PR

Melkiades: I think the problem comes from the following... How did you solve the problem @gmbecker?

> So to conclude, this isn't me not liking that we're doing linting, or even the settings of the linter. This is the linter fundamentally not functioning properly within the GH action.


_Originally posted by @gmbecker in https://github.com/insightsengineering/formatters/issues/83#issuecomment-1379467728_

gmbecker (approved):

PR #88, opened 2023-01-30
adding keyword internal, 

### LINKED ISSUE #55: Add missing topics to `_pkgdown.yml`
**Summary**

The following error was produced by the documentation GitHub actions workflow:

```r
! All topics must be included in reference index
✖ Missing topics: basic_matrix_form, basic_pagdf, default_hsep, font_lcpi, mf_strings, page_lcpp, page_types, table_inset, wrap_string
ℹ Either add to _pkgdown.yml or use @keyword internal
```

These topics need to be added to the __pkgdown.yml_ configuration.

shajoezhu:
close #55

### DIFF
--- a/R/matrix_form.R
+++ b/R/matrix_form.R
@@ -459,8 +459,8 @@ basic_matrix_form <- function(df, parent_path = "root") {
     exts <- rep(1L, NROW(df))
   }
   rowdf <- basic_pagdf(row.names(df),
-    extents = exts,
-    parent_path = parent_path
+                       extents = exts,
+                       parent_path = parent_path
   )
   formats <- cbind(
     "",
--- a/R/matrix_form.R
+++ b/R/matrix_form.R
@@ -52,10 +52,8 @@ mform_handle_newlines <- function(matform) {
     if (has_topleft) {
       newtl <- unlist(strsplit(tl, "\n"))
       if (length(newtl) > new_nlines_hdr) {
-        stop(
-          "Expanding top-left material resulted in more lines (", length(newtl),
-          "than fit in the header." # nocov
-        )
+        stop("Expanding top-left material resulted in more lines (", length(newtl), # nocov
+             "than fit in the header.") # nocov
       }
       newstrmat[1:new_nlines_hdr, 1] <- c(newtl, rep("", new_nlines_hdr - length(newtl)))
       newfrmmat[1:new_nlines_hdr, 1] <- "xx"
--- a/R/matrix_form.R
+++ b/R/matrix_form.R
@@ -256,6 +256,7 @@ matrix_print_form <- MatrixPrintForm
 #' @export
 #' @rdname mpf_accessors
 mf_strings <- function(mf) mf$strings
+
 #' @export
 #' @rdname mpf_accessors
 
--- a/R/matrix_form.R
+++ b/R/matrix_form.R
@@ -25,10 +25,10 @@ mform_handle_newlines <- function(matform) {
     hdr_inds <- 1:nr_header
     ## groundwork for sad haxx to get tl to not be messed up
     if (has_topleft) {
-      tl <- strmat[hdr_inds, 1]
-      strmat[hdr_inds, 1] <- ""
-      ## recalc them without topleft cause thats handled separately
-      row_nlines <- apply(strmat, 1, function(x) max(vapply(x, nlines, 1L), 1L))
+        tl <- strmat[hdr_inds, 1]
+        strmat[hdr_inds, 1] <- ""
+        ## recalc them without topleft cause thats handled separately
+        row_nlines <- apply(strmat, 1, function(x) max(vapply(x, nlines, 1L), 1L))
     } else {
       tl <- character()
     }
--- a/R/matrix_form.R
+++ b/R/matrix_form.R
@@ -25,10 +25,10 @@ mform_handle_newlines <- function(matform) {
     hdr_inds <- 1:nr_header
     ## groundwork for sad haxx to get tl to not be messed up
     if (has_topleft) {
-        tl <- strmat[hdr_inds, 1]
-        strmat[hdr_inds,1] <- ""
-        ## recalc them without topleft cause thats handled separately
-        row_nlines <- apply(strmat, 1, function(x) max(vapply(x, nlines, 1L), 1L))
+      tl <- strmat[hdr_inds, 1]
+      strmat[hdr_inds, 1] <- ""
+      ## recalc them without topleft cause thats handled separately
+      row_nlines <- apply(strmat, 1, function(x) max(vapply(x, nlines, 1L), 1L))
     } else {
       tl <- character()
     }
@@ -54,7 +54,7 @@ mform_handle_newlines <- function(matform) {
       if (length(newtl) > new_nlines_hdr) {
         stop(
           "Expanding top-left material resulted in more lines (", length(newtl),
-          "than fit in the header." #nocov
+          "than fit in the header." # nocov
         )
       }
       newstrmat[1:new_nlines_hdr, 1] <- c(newtl, rep("", new_nlines_hdr - length(newtl)))
@@ -459,8 +459,8 @@ basic_matrix_form <- function(df, parent_path = "root") {
     exts <- rep(1L, NROW(df))
   }
   rowdf <- basic_pagdf(row.names(df),
-                       extents = exts,
-                       parent_path = parent_path
+    extents = exts,
+    parent_path = parent_path
   )
   formats <- cbind(
     "",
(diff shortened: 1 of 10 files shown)


R/page_size.R
```diff
@@ -48,6 +48,7 @@ pg_dim_names <- list(
 #' @return for `page_types` a character vector of supported page types,
 #' for `page_dim` the dimensions (width, then height) of the selected page type.
 #'
+#' @keyword internal
 #' @export
```
Melkiades, line 52: One should be selected

Melkiades: Same for other internal calls

shajoezhu: Thanks man, I should have made this a draft PR, it was working progress, I realize the problem yesterday when I was making changes. let me get back to you on this. I think a few of these functions can be made internal, won't have to be exported. let me chck that

R/tostring.R
```diff
@@ -31,6 +31,7 @@ d_hsep_factory <- function() {
 #' if in a locale that uses a UTF character set, otherwise an ASCII hyphen
 #' with a once-per-session warning.
 #'
+#' @keyword internal
```
gmbecker, line 34: Used by rtables (and, I assume without bothering ot check, rlistings) to get default horizontal separator across UTF and non-UTF locales (in fact thats its entire purpose), can't be made internal.

R/page_size.R
```diff
@@ -86,6 +87,7 @@ page_dim <- function(page_type) {
 #' @return named list with `cpi` and `lpi`, the characters and lines per
 #' inch, respectively.
 #'
+#' @keyword internal
```
gmbecker, line 90: used in `rtables::export_as_pdf` and `rtables::export_as_txt`, must be exported until (at least) export logic is fully migrated to formatters

R/page_size.R
gmbecker: I don't think this should be internal because it is how the user can find out what page types we support.

R/tostring.R
```diff
@@ -367,6 +368,7 @@ new_line_warning <- function(str_v) {
 #' @return A string (`wrap_string` or character vector (`wrap_txt`) containing
 #'   the hard or soft word-wrapped content.
 #'
+#' @keyword internal
```
gmbecker, line 371: Not sure what the value of not exporting this would be, really, but I suppose it might be possible (doesn't seem to be called directly by rtables in code or tests)

R/page_size.R
```diff
@@ -136,6 +138,7 @@ font_lcpi <- function(font_family = "Courier", font_size = 12, lineheight = 1) {
 #'
 #' @return a named list containing `lpp` and `cpp` elements suitable
 #' for use by the pagination machinery.
+#' @keyword internal
```
gmbecker, line 141: Used in `rtables::paginate_table`, can't be (easily) made internal

R/matrix_form.R
```diff
@@ -7,6 +7,7 @@
 #' @rdname MatrixPrintForm
 #' @aliases MatrixPrintForm-class
 #' @exportClass MatrixPrintForm
+#' @keyword internal
```
gmbecker, line 10: This constructor is explicitly designed to be used by other packages (e.g., rtables, rlistings), and thus can never be made internal.

_pkgdown.yml
```diff
@@ -65,18 +67,13 @@ reference:
...
+  - title: Export
+    desc: Export into different object
+    contents:
+      - mpf_to_rtf
   - title: Miscellany
     desc: Other documented functions not interesting to end users
     contents:
       - print,ANY-method
       - ifnotlen0
-      - basic_matrix_form
```
gmbecker, line 73: (Most of) these changes will need to be reverted because the functions should be documented

gmbecker (changes requested):

R/matrix_form.R
shajoezhu, line 10:
```suggestion
```

R/page_size.R
```diff
@@ -48,6 +48,7 @@ pg_dim_names <- list(
 #' @return for `page_types` a character vector of supported page types,
 #' for `page_dim` the dimensions (width, then height) of the selected page type.
 #'
+#' @keyword internal
```
shajoezhu, line 51:
```suggestion
```

R/page_size.R
shajoezhu, line 90:
```suggestion
```

R/page_size.R
shajoezhu, line 141:
```suggestion
```

R/tostring.R
shajoezhu, line 34:
```suggestion
```

_pkgdown.yml
shajoezhu: Thanks @gmbecker , I am trying to familiar myself with the code base by trying to take some small issues at the moment :p

R/tostring.R
shajoezhu: I guess if users are not expecting to use this function, it should be internal, and not exported?

gmbecker: That rule of thumb is often true generally, but isn't particulatly true  in a package like `formatters`. Most of what it exports is intended for _package developers_ who are using the framework (currently, this means the developers of `rtables` and `rlistings`). Many things are exported by `formatters` that should never be called by end users but cannot be internal to `formatters`.

`wrap_txt` in point of fact, is used in a function that is currently in `rlistings` (`vec_nlines`). That could b emigrated, and might even be better in `formatters`, but the payoff would not be worth the changes (even though they'd be fairly minimal) at this stage because `rlistings` is the only thing that needs it.

We could make `wrap_string` internal while keeping `wrap_txt` exported, but that would be pretty strange in my opinion, so I'm loathe to do that.

formatters.Rproj
```diff
@@ -18,3 +18,4 @@ StripTrailingWhitespace: Yes
 BuildType: Package
 PackageUseDevtools: Yes
 PackageInstallArgs: --no-multiarch --with-keep.source
+PackageRoxygenize: rd,collate,namespace,vignette
```
Melkiades, line 21: is the vignette actually to be done here? documenting is producing files now whit this. let me remove the vignette. (the following comes from `tern`:
`PackageRoxygenize: rd,collate,namespace`)

shajoezhu: we can keep this right? there is https://github.com/insightsengineering/formatters/blob/main/vignettes/formatters.Rmd

shajoezhu: ah. sure. we can remove it then

R/tostring.R
```diff
@@ -152,8 +152,10 @@ setMethod("toString", "MatrixPrintForm", function(x,
   # Check for which row has indent
   ind_from_strings <- nchar(old_indent)[-seq_len(nlh)] > 0
   if (!all(ind_from_strings == ind_from_mf)) {
-    stop("Row-info and string indentations are different.", # nocov
-         " Please contact the maintainer, this should not happen.") # nocov
+    stop(
+      "Row-info and string indentations are different.", # nocov
+      " Please contact the maintainer, this should not happen."
+    ) # nocov
```
Melkiades, line 158: @gmbecker should I revert all style-related changes? some were automatic imo. The ones about export I did because I think it is more readable if internal/function-related-info stays close to the function definition while the rest is user-facing documentation and should be together, above. It is a change in style, but not in the final docs. I agree anyway that having a uniform style is best. If you want I can revert also these ones.

Melkiades (approved):
@gmbecker now it should be good to go

shajoezhu:
Hi @gmbecker , can you help to review this please. thanks

R/tostring.R
```diff
@@ -152,8 +151,10 @@ setMethod("toString", "MatrixPrintForm", function(x,
   # Check for which row has indent
   ind_from_strings <- nchar(old_indent)[-seq_len(nlh)] > 0
   if (!all(ind_from_strings == ind_from_mf)) {
-    stop("Row-info and string indentations are different.", # nocov
-         " Please contact the maintainer, this should not happen.") # nocov
+    stop(
```
gmbecker, line 154: This changes nocov behavior (and is an example of why we can't trust automatic code transformation to do the right thing in every single circumstance).

Previously, the full stop call was covered under #nocov, but after the change it isn't.

R/tostring.R
```diff
@@ -198,8 +203,10 @@ setMethod("toString", "MatrixPrintForm", function(x,
 
   # Additional safety check
   if (length(new_indent) > 0 && !all(nzchar(new_indent))) {
-    stop("Recovered indentation contains empty strings. This is an", # nocov
-         " indexing problem, please contact the maintainer, this should not happen.") # nocov
+    stop(
+      "Recovered indentation contains empty strings. This is an", # nocov
```
gmbecker, line 207: As above, nocov behavior has been changed again here.

gmbecker (changes requested):

R/tostring.R
gmbecker: fyi @cicdguy just so you're generally aware of this issue with automated styling.

gmbecker (approved):

PR #91, opened 2023-01-30
adding check and fixing indexing error

### LINKED ISSUE #90: Fix bug when there is grouping in header (e.g. with `\n`)
Ref: https://github.com/Roche/rtables/issues/510

Melkiades:
Fixes #90

### DIFF
--- a/R/tostring.R
+++ b/R/tostring.R
@@ -106,7 +106,7 @@ setMethod("toString", "MatrixPrintForm", function(x,
                                                   hsep = default_hsep()) {
   mat <- matrix_form(x, TRUE)
   inset <- table_inset(mat)
-
+  browser()
   if (is.null(widths)) {
     widths <- propose_column_widths(x)
   }
@@ -130,9 +130,22 @@ setMethod("toString", "MatrixPrintForm", function(x,
   cell_widths_mat <- .calc_cell_widths(mat, widths, col_gap)
   ## wrap_string calls strwrap, which destroys whitespace so we need to make
   ## sure to put the indents back in
+
+  # See if indentation is properly set
+  inner_tbl_indent <- mf_rinfo(mat)$indent
+  nlh <- mf_nlheader(mat)
   old_indent <- gsub("^([[:space:]]*).*", "\\1", mat$strings[, 1])
   need_reindent <- nzchar(old_indent)
+  # Check for which row has indent
+  ind_from_strings <- as.integer(need_reindent)[-seq_len(nlh)]
+  ind_from_mf <- inner_tbl_indent
+  if (!checkmate::check_set_equal(ind_from_strings, ind_from_mf)) {
+    stop("Row-info and string indentations are different.", # nocov
+         "Please contact the maintainer, this should not happen.") # nocov
+  }
+
   reindent_old_idx <- mf_lgrouping(mat)[need_reindent]
+  # Taking care in advance of indented word wrappings
   cell_widths_mat[need_reindent, 1] <- cell_widths_mat[need_reindent, 1] - nchar(old_indent)[need_reindent]
 
   new_strings <- matrix(unlist(mapply(wrap_string,
@@ -151,12 +164,12 @@ setMethod("toString", "MatrixPrintForm", function(x,
   if (anyNA(reindent_new_idx)) {
     stop(
       "Unable to remap indenting after cell content text wrapping. ", # nocov
-      "Please contact the maintainer, this should not happen"
+      "Please contact the maintainer, this should not happen."
     ) # nocov
   }
   # Adding the indentation back in
-  new_indent <- old_indent[mf_lgrouping(mat)[reindent_new_idx]]
-  mat$strings[which(reindent_new_idx), 1] <- paste0(
+  new_indent <- old_indent[reindent_new_idx]
+  mat$strings[reindent_new_idx, 1] <- paste0(
     new_indent,
     mat$strings[reindent_new_idx, 1]
   )


gmbecker:
@cicdguy we are still getting the (quite disruptive) false positives fromt he SuperLinter for the formatters package.

R/tostring.R
```diff
@@ -130,9 +130,22 @@ setMethod("toString", "MatrixPrintForm", function(x,
   cell_widths_mat <- .calc_cell_widths(mat, widths, col_gap)
   ## wrap_string calls strwrap, which destroys whitespace so we need to make
   ## sure to put the indents back in
+
+  # See if indentation is properly set
+  ind_from_mf <- mf_rinfo(mat)$indent > 0
+  nlh <- mf_nlheader(mat)
   old_indent <- gsub("^([[:space:]]*).*", "\\1", mat$strings[, 1])
```
gmbecker, line 137: We can't rely on this as pointed out by the related bug you filed. We should be using `ind_from_mf` throughout instead of `old_indent` (the way it was previously calculated)

gmbecker (changes requested):

R/tostring.R
Melkiades: Yes, this is causing the latest error I filed. I am working on it, bc indent does not keep track of how many spaces I think and I am searching what does

Melkiades:
> > @cicdguy we are still getting the (quite disruptive) false positives fromt he SuperLinter for the formatters package.
> 
> I believe [cb2041e](https://github.com/insightsengineering/formatters/pull/91/commits/cb2041ebcb52fd14994eeff082ae5483990f220b) has resolved it. Are you talking about the annotations? If so, I can disable those also.

I think Gabe was referring to the error there was before. Thank you for fixing it ;)

R/matrix_form.R
```diff
@@ -51,14 +55,14 @@ mform_handle_newlines <- function(matform) {
         )
       }
       newstrmat[1:new_nlines_hdr, 1] <- c(tl, rep("", new_nlines_hdr - length(tl)))
-      newfrmmat[1:new_nlines_hdr, 1] <- "xx"
+      newfrmmat[1:new_nlines_hdr, 1] <- "xx" # Why is this needed?
```
Melkiades, line 58: @gmbecker Is it needed?

R/matrix_form.R
```diff
@@ -51,14 +55,14 @@ mform_handle_newlines <- function(matform) {
...
     }
     mf_strings(matform) <- newstrmat
     mf_formats(matform) <- newfrmmat
     mf_spans(matform) <- expand_mat_rows(mf_spans(matform), row_nlines, rep_vec_to_len)
     mf_aligns(matform) <- expand_mat_rows(mf_aligns(matform), row_nlines, rep_vec_to_len)
     mf_display(matform) <- expand_mat_rows(mf_display(matform), row_nlines, rep_vec_to_len)
-    mf_lgrouping(matform) <- rep(1:nrows, times = row_nlines)
+    mf_lgrouping(matform) <- rep(mf_lgrouping(matform), times = row_nlines)
```
Melkiades, line 65: This was a major bug for `\n` in cols when `top_left` material was present

R/tostring.R
```diff
@@ -103,24 +105,78 @@ setMethod("toString", "MatrixPrintForm", function(x,
...
+                                                  hsep = default_hsep(),
+                                                  indent_size = NULL) {
+
+  checkmate::assert_integerish(indent_size, null.ok = TRUE)
+  checkmate::assert_flag(tf_wrap)
+
+  mat <- matrix_form(x, indent_rownames = TRUE) # Fix for indent_size should be here
   inset <- table_inset(mat)
 
+  # Standard indentation size and inserted one
```
Melkiades, line 117: The following is a tentative reparation for the case in which indent_size is set only here and not before. There is something about widths and ncchar that is not calculated correctly afterward, so even if the indentation is reconstructed correctly it produces a wider table at the end. See related test in rtables.

R/tostring.R
```diff
@@ -103,24 +105,78 @@ setMethod("toString", "MatrixPrintForm", function(x,
...
+      # formats
+      mat$formats[-seq_len(mf_nlheader(mat)), 1] <-
+        paste(
+          sapply(mf_rinfo(mat)$indent, function(i) {
+            paste0(rep(" ", i * diff_indent_size), collapse = "")
+          }),
+          mat$formats[-seq_len(mf_nlheader(mat)), 1]
+        )
+    } else {
+      # Take out a space if inserted indentation is smaller
```
Melkiades, line 139: not working as expected. Still to correct for negative cases

gmbecker: Negative indents have never been supported, minimum indent is 0. We should probably catch such things as an error in various places but that should be filed as a separate issue.

R/tostring.R
gmbecker: indent size shouldn't be set for the `toString MatrixPrintForm` method, that should always use the indent_size on the MatrixPrintForm object. indent_size is set (on the `MatrixPrintForm` object) for the `toString VTableTree` method in rtables, because rtables doesn't have a concept of indent size in the object (as that is an aspect of rendering). Sorry that was not clear from my previous comment.

R/tostring.R
```diff
@@ -130,33 +186,83 @@ setMethod("toString", "MatrixPrintForm", function(x,
...
+  # Additional safety check
+  if (length(new_indent) > 0 && !all(nzchar(new_indent))) {
+    stop("Recovered indentation contains empty strings. This is an", # nocov
+         " indexing problem, please contact the maintainer, this should not happen.") # nocov
+  }
+
+  # Indentation is different for topleft material
+  if (isTRUE(mf_has_topleft(mat))) {
+    # mf_nlheader counts actual header lines while mf_nrheader is 'virtual'
+    # A bit of an hack, we should see what happens if there are \n in topleft
```
gmbecker, line 258: \n is _never_ allowed in top-left, and isn't going to be. They can specify each single line topleft part separately as a vector but there should not ever be new lines there. 

The only timme I can think of that its even possible that could happen is if newlines are getting variable labels and the variable labels have newlines in them, but we can just tell them not to do that, or hard split it in the actual topleft setter. By the time something is in the topleft slot it should *not* have newlines.

R/matrix_form.R
gmbecker: This is setting the "format" for the topleft area to `"xx"` (mostly so that it cannot possibly have a different format, I think). I agree that its odd but its unrelated to the work at hand so lets leave it for now and we an do a deeper investigation once the fire is put out.

gmbecker:
> > @cicdguy we are still getting the (quite disruptive) false positives fromt he SuperLinter for the formatters package.
> 
> I believe [cb2041e](https://github.com/insightsengineering/formatters/pull/91/commits/cb2041ebcb52fd14994eeff082ae5483990f220b) has resolved it. Are you talking about the annotations? If so, I can disable those also.

@cicdguy Yes please do, they're very disruptive and completely useless when trying to read the changes  in the face of the mass of false positives

R/matrix_form.R
```diff
@@ -212,7 +219,8 @@ MatrixPrintForm <- function(strings = NULL,
       prov_footer = prov_footer,
       col_gap = col_gap,
       table_inset = as.integer(table_inset),
-      has_topleft = has_topleft
+      has_topleft = has_topleft,
+      indent_size = indent_size
```
Melkiades, line 223: I do not understand if this is good to be added here... I do not know if this is followed by an actual indentation somewhere before. Afterwards, it is for sure reconstructed as such

R/tostring.R
Melkiades: Thank you! Now it is clear. I got confused by the fact that `toString` is both in rtables and formatters eheh. It makes sense now! Reverted these changes

R/tostring.R
Melkiades: eheh yes, this was a sort of relative indentation. It does not make a ton of sense now. I will add the check for indent_size > 0

R/tostring.R
Melkiades: Yes, I fully agree. Having \n there makes things very complicated without any actual gain. I do not envision people starting to do this too soon. We should add an error for it (another issue I guess).

Melkiades: Modified comments accordingly

shajoezhu (approved):
lgtm! Thanks a lot @Melkiades ! tested downstream in Chevron, it is now working.

shajoezhu:
Hi @gmbecker , hope you are well. Can you accept the changes here. The problem is now resolved. and we have several PRs downstream in Chevron is blocked by this. Thanks a lot!

PR #97, opened 2023-02-03
Migrate exporter functions to `formatters`

### LINKED ISSUE #96: Migrate Exporter functions into `formatters`
this is the counterpart rtables issue for https://github.com/Roche/rtables/issues/491, and is currently blocking https://github.com/insightsengineering/rlistings/issues/61

- [ ] text-file exporter export_as_txt

Will revivit this issue once priority is updated for the following items. 
- flextable exporter 
- PDF export_as_pdf
- RTF export_as_rtf

edelarua:
Closes #96 and https://github.com/insightsengineering/sme-tasks/issues/578

`rtables` counterpart: https://github.com/Roche/rtables/pull/519


edelarua:
Errors come from using CRAN version of `rtables` which is out of date so does not include additional arguments (i.e. arguments in `analyze` and `paginate_table`) used in exporters & their tests. Consequently Roxygen check automatically removes documentation of these inherited arguments.

gmbecker (changes requested):
These functions can't juts be directly transplanted. There are things that are rtables specific that need to be refactored to only operate on the `MatrixPrintForm`, and there may need to be changes (additions) to `MatrixPrintForm` to accomplish that. (if that weren't the case I would have just done it instead of filing an issue because it would have been easier).

We need to take a careful look at what these exporters are doing, take note of everythhning they do that isn't based completely on the MPF, and then figure out how to rework the logic so that it can be based on the MPF.

Also there can't be any `rtables::` in formatters as that would be a circular dependency.

edelarua:
> These functions can't juts be directly transplanted. There are things that are rtables specific that need to be refactored to only operate on the `MatrixPrintForm`, and there may need to be changes (additions) to `MatrixPrintForm` to accomplish that. (if that weren't the case I would have just done it instead of filing an issue because it would have been easier).
> 
> We need to take a careful look at what these exporters are doing, take note of everythhning they do that isn't based completely on the MPF, and then figure out how to rework the logic so that it can be based on the MPF.
> 
> Also there can't be any `rtables::` in formatters as that would be a circular dependency.

Thanks @gmbecker. Joe wants to prioritize this issue but I don't think he realized that it would create circular dependencies and  be complex/require additional refactoring. I will close these PRs for now and flag this with him.

shajoezhu:
Hi @edelarua , sorry, my bad, I should have given more input/instruction on the issue. This is a relatively large task, will definitely work on both in formatters and rtables repo, to split the functions and move the parts that have not dependcies within rtables into formatters, and then refactor the functions. let's work though this in the coming week. Thanks a lot

PR #101, opened 2023-02-10
Update check as-cran and default landing page

walkowif:
Solves these two issues:
https://github.com/insightsengineering/idr-tasks/issues/477
https://github.com/insightsengineering/idr-tasks/issues/478


shajoezhu (approved):
approve to unblock merge

PR #112, opened 2023-03-02
State initially submitted to cran as 0.4.0

gmbecker:
[skip vbump]

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: formatters
 Title: ASCII Formatting for Values and Tables
-Version: 0.3.4.22
+Version: 0.4.0
 Date: 2023-03-01
 Authors@R: c(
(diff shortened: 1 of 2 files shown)


gmbecker:
We are going to have to just ride out the spell check failure for this release, as the package has already been submitted (there are only 2 words, both in the NEWS.md file)

shajoezhu:
Hi @gmbecker , I was wondering why we can't fix it here. and then submit to cran? have you submitted already? then updating the file here?

gmbecker:
> Hi @gmbecker , I was wondering why we can't fix it here. and then submit to cran? have you submitted already? then updating the file here?

Yes, I have submitted already.

shajoezhu:
Hi @gmbecker , i think it's ok to fix the spelling here and merge it into main. since it is already on cran

PR #151, opened 2023-05-16
release 0.5.0

Melkiades:
[skip vbump] [skip spelling]

#149

PR #153, opened 2023-05-16
update news and bump version

gmbecker:


shajoezhu:
hi @gmbecker , can you review @Melkiades 's PR #151 instead. we need to skip vbump. Thanks

gmbecker:
@shajoezhu no. because I did this so that I can propogate the correct version dependencies to rtables and rlistings BEFORE we encapsulate things for final release, because we're not sure we're ready yet.

Issue #143, opened 2023-05-05
Missing unit test report


gmbecker:
when doing `R CMD check --as-cran`:

```
* checking CRAN incoming feasibility ... NOTE
Maintainer: ‘Gabriel Becker <gabembecker@gmail.com>’

Version contains large components (0.4.1.9001)

Found the following (possibly) invalid URLs:
  URL: https://insightsengineering.github.io/formatters/main/unit-test-report/
    From: README.md
    Status: 404
    Message: Not Found
```

gmbecker:
@cicdguy ping on this. We will need it fixed before CRAN release

shajoezhu:
pinging @walkowif on this.

walkowif:
This is caused by the fact that [this step](https://github.com/insightsengineering/r.pkg.template/blob/main/.github/workflows/build-check-install.yaml#L288-L304) in the R CMD check workflow expects the file to be named `tests/testthat.R` in order to configure the JUnit reporter to save the result to `junit-result.xml`.
Instead of `tests/testthat.R`, there's `tests/run_tests.R` in the `formatters` repository.
This way, `junit-result.xml` is not generated and the unit test report is missing from the documentation.

There are two possibilities:
* renaming `tests/run_tests.R` to `tests/testthat.R`,
* or modifying the pkgdown configuration not to include the link to unit test report ([these two lines](https://github.com/insightsengineering/formatters/blob/main/_pkgdown.yml#L14C31-L15) will have to be removed)

PR #158, opened 2023-05-23
rename test file to testthat.R

gmbecker:
[skip vbump]


gmbecker:
@walkowif @cicdguy Why did renaming the test running file (as suggested in #143 ) cause this to suddenly fail when it was passing check before (and still does for me locally). Any ideas?

walkowif:
The R CMD check workflow modifies the `tests/testthat.R` so that it contain [these lines](https://github.com/insightsengineering/r.pkg.template/blob/main/.github/workflows/build-check-install.yaml#L292-L300) (but only if this file exists, that's why the file name change was suggested).
After searching for the error appearing in the workflow (`no applicable method for 'xml_add_child' applied to an object of class "NULL"`) it looks like it may be a result of incorrectly placed `context()` in the tests.

gmbecker:
@walkowif fyi, as far as I know we are using contexts correctly there, and contexts are only deprecated in testthat "third-edition", which formatters and rtables don't use. I can try removing them here, and see if it works without changing editions.

if testthat 3e is actually _required_ to use the idr automation, that is something that should be well documented (since AFAIK these workflows are public)

PR #163, opened 2023-06-07
vbump deps

pawelru:
WIP

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -5,6 +5,9 @@ on:
   schedule:
     - cron: '45 3 * * 0'
   workflow_dispatch:
+  push:
+    branches:
+      - min_deps
 
 jobs:
   dependency-test:
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -9,7 +9,6 @@ on:
     branches:
       - min_deps
 
-
 jobs:
   dependency-test:
     strategy:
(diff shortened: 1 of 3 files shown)


.github/workflows/integration-tests.yaml
```diff
@@ -5,6 +5,9 @@ on:
   schedule:
     - cron: '45 3 * * 0'
   workflow_dispatch:
+  push:
+    branches:
+      - min_deps
```
pawelru, line 10: Need to revert it once completed

```suggestion
```

.github/workflows/integration-tests.yaml
```diff
@@ -5,6 +5,9 @@ on:
...
```
pawelru, line 11:
```suggestion


```

.github/workflows/integration-tests.yaml
```diff
@@ -5,6 +5,10 @@ on:
...
+
```
pawelru, line 11:
```suggestion
```

gmbecker (changes requested):
Do we have evidence that versions earlier than these cause problems? If not, this ends up inpractice imposing a pretty strict modernity requirement on rtables which is at odds with the way production systems tend to move and be updated slowly.

gmbecker (changes requested):
Do we have evidence that versions earlier than these cause problems? If not, this ends up inpractice imposing a pretty strict modernity requirement on rtables which is at odds with the way production systems tend to move and be updated slowly.

PR #171, opened 2023-07-07
Add test for `default_hsep`

### LINKED ISSUE #154: Missing test for function "default_hsep"
See at https://github.com/insightsengineering/formatters/blob/7ae3f0f14e008cc118b2ed0778d3a5af67b16215/R/tostring.R#L28

autovalidate tests flagged

```
 - requirement #0.03 'Default horizontal Separator' is not covered by any test
```

edelarua:
Add test to pass autovalidate check.

Closes #154

### DIFF
--- a/tests/testthat/test-formatters.R
+++ b/tests/testthat/test-formatters.R
@@ -537,3 +537,7 @@ mpf <- MatrixPrintForm(strings = strs, spans = spans, aligns = aligns,
 
 expect_equal(length(grep("spn_val", toString(mpf))),
              1L)
+
+test_that("default_hsep works as expected", {
+  expect_identical(default_hsep(), "—")
+})


tests/testthat/test-formatters.R
```diff
@@ -537,3 +537,7 @@ mpf <- MatrixPrintForm(strings = strs, spans = spans, aligns = aligns,
 
 expect_equal(length(grep("spn_val", toString(mpf))),
              1L)
+
+test_that("default_hsep works as expected", {
+  expect_identical(default_hsep(), "—")
```
gmbecker, line 542: This will fail on old systems which do not fully support UTF-8 (at least one of CRAN's buidl machines was such a system in the last cycle).

Also, this is a false positive. we have tests for `export_as_txt`, which calls default_hsep by default, so this actually is covered...

gmbecker (changes requested):

tests/testthat/test-formatters.R
gmbecker: fyi @pawelru @shajoezhu more false positives from autovalidate.

shajoezhu: thanks @gmbecker , hi @dgkf-roche fyi. Was wondering will autovalidater be updated?

PR #174, opened 2023-07-11
Update `export_as_txt` to print split label when using `page_by` with one level

edelarua:
Closes https://github.com/insightsengineering/rtables/issues/653

### DIFF
--- a/R/pagination.R
+++ b/R/pagination.R
@@ -889,6 +889,10 @@ paginate_indices <- function(obj,
          pag_col_indices = pag_col_indices)
 }
 
+setGeneric("has_page_title", function(obj) standardGeneric("has_page_title"))
+
+setMethod("has_page_title", "ANY", function(obj) length(page_titles(obj)) > 0)
+
 #' @rdname paginate_indices
 #' @export
 paginate_to_mpfs <- function(obj,
@@ -958,7 +962,7 @@ paginate_to_mpfs <- function(obj,
                       nosplitin = nosplitin,
                       verbose = verbose)
         return(unlist(deep_pag, recursive = FALSE))
-    } else if (!is.null(attr(obj, "page_title_prefix")) && !is.na(attr(obj, "page_title_prefix"))) {
+    } else if (has_page_title(fpags[[1]])) {
       obj <- fpags[[1]]
     }
 
--- a/R/pagination.R
+++ b/R/pagination.R
@@ -958,6 +958,8 @@ paginate_to_mpfs <- function(obj,
                       nosplitin = nosplitin,
                       verbose = verbose)
         return(unlist(deep_pag, recursive = FALSE))
+    } else if (!is.null(attr(obj, "page_title_prefix")) && !is.na(attr(obj, "page_title_prefix"))) {
+      obj <- fpags[[1]]
     }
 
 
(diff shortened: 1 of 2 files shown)


R/pagination.R
```diff
@@ -958,6 +958,8 @@ paginate_to_mpfs <- function(obj,
                       nosplitin = nosplitin,
                       verbose = verbose)
         return(unlist(deep_pag, recursive = FALSE))
+    } else if (!is.null(attr(obj, "page_title_prefix")) && !is.na(attr(obj, "page_title_prefix"))) {
```
edelarua, line 961: @gmbecker in `rtables` there are functions `ptitle_prefix` and `has_force_pag` which are used to do these checks (and are used in `paginate_table`). Should these functions be moved to `formatters` so that they can be used here as well?

gmbecker: so prtitle_prefix is what the parent carries around that gets inserted into the page titles, but we should be able to just check `page_title(fpags[[1]])` to see if stuff is in there. 

Please confirm this with tests of both cases @edelarua

gmbecker: To explain further, `has_force_pag` is to detect whether forced pagination is needed. So we could refactor how `paginate_to_mpfs` works to check explicitly, but as it is now we are trying to determine whether it happened *after* it occurs. 

We should probably wrap the check I suggested above in something like `has_page_title`, or something

edelarua:
Tests added here: https://github.com/insightsengineering/rtables/pull/681

shajoezhu (approved):
lgtm! many thanks @edelarua , it works well. and is getting expected output

```
> basic_table() %>%
+     split_cols_by("ARM") %>%
+     add_colcounts() %>%
+     add_overall_col("overall") %>%
+     split_rows_by("PARAMCD", labels_var = "PARAM",split_label = "aaa", label_pos = "topleft",split_fun = drop_split_levels, page_by = T) %>%
+     split_rows_by("AVISIT", label_pos = "topleft",split_fun = drop_split_levels) %>%
+     summarize_vars(
+         c("AVAL", "CHG"),
+         .stats = c("n", "mean_sd", "median", "range", "count_fraction")
+     ) %>%
+     append_topleft(paste0("  asd adsa")) %>%
+     build_table(ex_adlb %>% filter(PARAMCD == "ALT")) %>%
+     export_as_txt(lpp = 100) %>%
+     cat()

aaa: Alanine Aminotransferase Measurement

——————————————————————————————————————————————————————————————————————————————————————————————————
aaa                                                                                               
  AVISIT                                A: Drug X      B: Placebo    C: Combination     overall   
  asd adsa                               (N=938)        (N=938)         (N=924)         (N=2800)  
——————————————————————————————————————————————————————————————————————————————————————————————————
SCREENING                                                                                         
  AVAL                                                                                            
    n                                      134            134             132             400     
    Mean (SD)                           50.0 (7.2)     50.8 (8.4)      50.2 (7.6)      50.3 (7.8) 
    Median                                 49.7           50.1            49.6            49.8    
    Min - Max                          31.7 - 71.2    29.3 - 69.2     26.9 - 70.0     26.9 - 71.2 
  CHG                                                                                             
    n                                       0              0               0               0      
    Mean (SD)                               NA             NA              NA              NA     
    Median                                  NA             NA              NA              NA     
    Min - Max                               NA             NA              NA              NA     
```

PR #176, opened 2023-07-13
Create `obj_align`, `fmt_config` constructor/class

### LINKED ISSUE #82: `fmt_config` constructor and class to bundle format, na_str, and align instructions
Arose from insightsengineering/rlistings#72

edelarua:
Closes #82


tests/testthat/test-formatters.R
```diff
@@ -537,3 +537,15 @@ mpf <- MatrixPrintForm(strings = strs, spans = spans, aligns = aligns,
 
 expect_equal(length(grep("spn_val", toString(mpf))),
              1L)
+
+testthat::test_that("fmt_config works as expected", {
+  x <- fmt_config()
+  expect_identical(obj_format(x), NULL)
```
edelarua, line 543: Still need to figure out why this returns `\001NULL\001` instead of `NULL` (`x@format` returns correct result)

shajoezhu (approved):
lgtm! Thanks  @edelarua . UATed by @Teninq downstream as well. I think we can proceed with this

shajoezhu:
can you please update the NEWS as well. Thanks

tests/testthat/test-formatters.R
```diff
@@ -537,3 +537,15 @@ mpf <- MatrixPrintForm(strings = strs, spans = spans, aligns = aligns,
 
 expect_equal(length(grep("spn_val", toString(mpf))),
              1L)
+
+testthat::test_that("fmt_config works as expected", {
+  x <- fmt_config()
+  expect_identical(slot(x, "format"), NULL)
```
Melkiades, line 543: isnt the standard "xx" for format?

edelarua: I think the default is `NULL` since it has to generalize to both numeric and non-numeric formats. From what I can tell `format_value`, `analyze`, etc. all use `NULL` as default.

R/format_value.R
```diff
@@ -387,3 +387,28 @@ format_value <- function(x, format = NULL, output = c("ascii", "html"), na_str =
     txt
   }
 }
+
+setClassUnion("FormatSpec", c("NULL", "character", "function", "list"))
+setClassUnion("characterOrNULL", c("NULL", "character"))
+setClass("fmt_config",
+         slots = c(format = "FormatSpec", format_na_str = "characterOrNULL", align = "characterOrNULL"))
```
gmbecker, line 394: There's no reason to use characterOrNULL here, befcause format_na_str will have a default string value.

R/generics.R
```diff
@@ -318,6 +318,22 @@ setMethod("obj_na_str<-", "ANY", function(obj, value) {
   obj
 })
 
+#' @rdname lab_name
+#' @export
+setGeneric("obj_align", function(obj) standardGeneric("obj_align"))
+#' @rdname lab_name
+#' @exportMethod obj_align
+setMethod("obj_align", "ANY", function(obj) attr(obj, "align", exact = TRUE))
```
gmbecker, line 326: This shouldn't be using attr, these are S4 classes, so it should be `@`, need to get all the way out of the habit of accessing slots with attr.

R/generics.R
```diff
@@ -318,6 +318,22 @@ setMethod("obj_na_str<-", "ANY", function(obj, value) {
...
+
+#' @rdname lab_name
+#' @export
+setGeneric("obj_align<-", function(obj, value) standardGeneric("obj_align<-"))
+#' @exportMethod obj_align<-
+#' @rdname lab_name
+setMethod("obj_align<-", "ANY", function(obj, value) {
+  attr(obj, "align") <- value
```
gmbecker, line 334: the above comment about `@` is even much more true for setting. the fact that this happens to work is an implementation detail of S4 which should_never_ be relied upon.

tests/testthat/test-formatters.R
```diff
@@ -537,3 +537,15 @@ mpf <- MatrixPrintForm(strings = strs, spans = spans, aligns = aligns,
...
+  expect_identical(obj_na_str(x), "NA")
+  expect_identical(obj_align(x), "center")
+
+  x <- fmt_config(format = "xx.xx", na_str = "<Missing>", align = "right")
+  expect_identical(obj_format(x), "xx.xx")
+  expect_identical(obj_na_str(x), "<Missing>")
+  expect_identical(obj_align(x), "right")
```
gmbecker, line 550: There are no tests for the `obj_align` setter.

gmbecker (commented):
@shajoezhu this was not ready to be merged, especially on the day a release was supposed to occur

R/generics.R
shajoezhu: Thanks @gmbecker , we can rasise a new issue for the refactoring here.

R/generics.R
shajoezhu: Thanks @gmbecker , we can rasise a new issue for the refactoring here.

tests/testthat/test-formatters.R
shajoezhu: hi @gmbecker , how would you normally test for setters and getters?

Melkiades: done in https://github.com/insightsengineering/formatters/pull/179

R/generics.R
Melkiades: Done in https://github.com/insightsengineering/formatters/pull/179

R/format_value.R
Melkiades: now standards are all `NULL` to be set in `rlistings`: https://github.com/insightsengineering/formatters/pull/179

PR #177, opened 2023-07-13
updating description, news, change version to 0.5.1, [skip vbump]

shajoezhu:


clarkliming (approved):
unblock PR. already validated

PR #184, opened 2023-07-18
Reinstate defaults to `NULL`

Melkiades:
`formatters` should not set defaults beforehand on `fmt_config` as they will be different for `rtables` and `rlistings`. Setting them to `NULL` here will trigger the defaults in `rlistings` that make `align="left"` global default.

After this we will have to update the default snapshots in `chevron` @BFalquet @clarkliming and `scda.test` @shajoezhu. Whenever you want, we can go ;)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,8 +1,7 @@
 ## formatters 0.5.0.9006
  * New generic `getter` and `setter` for `align` (`obj_align` and `obj_align<-`).
  * New `fmt_config` class to bundle together `format`, `na_str`, and `align` instructions.
- * Reverting default for alignment (from `NULL` to `center`) and `na_str` 
-   from `NULL` to `"NA"`. This affects only `rlistings`, where the new default takes effect.
+ * Default for `fmt_config` alignment and `na_str` set to `NULL`. The default values are set in `rlistings`.
 
 ## formatters 0.5.0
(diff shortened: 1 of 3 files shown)


shajoezhu:
for now. I rather not to merge this PR into the release, the autovalidation takes a few hours to run, and the machinary is in place, and either rlistings or chevron calls can set the correct value when needs to. so that's ok.

gmbecker:
> the goal for "fmt_config" is to enable rlistings alignment and other settings. This feature is not well thought through for rtables yet. But I agree, this would be a nice placeholder in formatters, and potential nice use case in rtables, but we have not met such case yet.
> 
> Bottom line, whatever changes we make in formatters regarding this topic, is too ensure rlistings to work

Any changes to formatters must work for both rlistings and rtables. Formatters underlies them both, by design. Furthermore, if a default is going to reflect one of the desired defaults, it _**must**_ be rtables, as that system is much more important. If rlistings wants a different default than rtables (something I really don't think is worth the effort and was not on the list of features we originally specced out) then it is its job (ie rlistings' job) to implement that.

Melkiades:
After discussion, `formatters` NEEDS a default, i.e. the one for `rtables`. As it is possible to change the global formats from default_formatting, it is better to do it at that time. @BFalquet fyi

PR #192, opened 2023-08-23
Fix bug that threw warning when prov_footer len > 1 (rtables#705)

gmbecker:
fixes warning regarding prov footer mentioned in  insightsengineering/rtables#705


shajoezhu (approved):
lgmt! Thanks @gmbecker

PR #215, opened 2023-10-17
Add newline character support for ref footnotes

Melkiades:
(pagination does not count it though for now) needed for https://github.com/insightsengineering/rtables/issues/750

### DIFF
--- a/R/matrix_form.R
+++ b/R/matrix_form.R
@@ -15,6 +15,8 @@ mform_handle_newlines <- function(matform) {
   has_topleft <- mf_has_topleft(matform)
   strmat <- mf_strings(matform)
   frmmat <- mf_formats(matform)
+  spamat <- mf_spans(matform)
+  alimat <- mf_aligns(matform)
   nr_header <- mf_nrheader(matform)
   hdr_inds <- 1:nr_header
 
@@ -28,6 +30,15 @@ mform_handle_newlines <- function(matform) {
     tl_to_add_back <- strsplit(paste0(tl, collapse = "\n"), split = "\n", fixed = TRUE)[[1]]
     how_many_nl <- length(tl_to_add_back)
   }
+
+  # pre-proc in case of wrapping and \n
+  line_grouping <- mf_lgrouping(matform)
+  strmat <- .compress_mat(strmat, line_grouping, "nl")
+  frmmat <- .compress_mat(frmmat, line_grouping, "unique") # never not unique
+  spamat <- .compress_mat(spamat, line_grouping, "unique")
+  alimat <- .compress_mat(alimat, line_grouping, "unique")
+  line_grouping <- unique(line_grouping)
+
   # nlines detects if there is a newline character
   row_nlines <- apply(strmat, 1, function(x) max(vapply(x, nlines, 1L), 1L))
 
@@ -73,13 +84,35 @@ mform_handle_newlines <- function(matform) {
 
     mf_strings(matform) <- newstrmat
     mf_formats(matform) <- newfrmmat
-    mf_spans(matform) <- expand_mat_rows(mf_spans(matform), row_nlines, rep_vec_to_len)
-    mf_aligns(matform) <- expand_mat_rows(mf_aligns(matform), row_nlines, rep_vec_to_len)
+    mf_spans(matform) <- expand_mat_rows(spamat, row_nlines, rep_vec_to_len)
+    mf_aligns(matform) <- expand_mat_rows(alimat, row_nlines, rep_vec_to_len)
 ##    mf_display(matform) <- expand_mat_rows(mf_display(matform), row_nlines, rep_vec_to_len)
-    mf_lgrouping(matform) <- rep(mf_lgrouping(matform), times = row_nlines)
+    mf_lgrouping(matform) <- rep(line_grouping, times = row_nlines)
   }
 
   matform
+}
+
+# Helper function to recompact the lines following line groupings to then have them expanded again
+.compress_mat <- function(mat, line_grouping, collapse_method = c("nl", "unique")) {
+  list_compacted_mat <- lapply(unique(line_grouping), function(lg){
+    apply(mat, 2, function(mat_cols) {
+      col_vec <- mat_cols[which(line_grouping == lg)]
+      if (collapse_method[1] == "nl") {
+        paste0(col_vec, collapse = "\n")
+      } else {
+        val <- unique(col_vec)
+        val <- val[nzchar(val)]
+        if (length(val) > 1) {
+          stop("Problem in linegroupings! Some do not have the same values.") # nocov
+        } else if(length(val) < 1) {
+          val <- "" # Case in which it is only ""
+        }
+        val[[1]]
+      }
+    })
+  })
+  do.call("rbind", list_compacted_mat)
 }
 
 disp_from_spans <- function(spans) {
--- a/R/matrix_form.R
+++ b/R/matrix_form.R
@@ -24,13 +24,16 @@ mform_handle_newlines <- function(matform) {
     # Header indices
     hdr_inds <- 1:nr_header
 
+    # Padder should be bottom aligned if no topleft (case of rlistings)
+    tl_padder <- ifelse(has_topleft, pad_vert_top, pad_vert_bottom)
+
     ## used below even though we don't store it on the resulting object
     new_nlines_hdr <- sum(row_nlines[hdr_inds])
     newstrmat <- rbind(
       cbind(
         expand_mat_rows(strmat[hdr_inds, 1, drop = FALSE],
                         row_nlines[hdr_inds],
-                        cpadder = pad_vert_top # topleft info is top aligned
+                        cpadder = tl_padder # topleft info is top aligned
         ),
         expand_mat_rows(strmat[hdr_inds, -1, drop = FALSE],
                         row_nlines[hdr_inds],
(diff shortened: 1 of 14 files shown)


R/pagination.R
```diff
@@ -992,21 +992,29 @@ paginate_to_mpfs <- function(obj,
...
-      pagmats <- lapply(page_indices$pag_row_indices,
-                        function(ii) matrix_form(obj[ii, ], TRUE, TRUE,
-                                                 indent_size = indent_size))
+    # This needs to be fixed better
+    if (inherits(obj[[1]], "listing_df") && length(obj) < 2) {
+      pagmats <- lapply(
+        page_indices$pag_row_indices,
+        function(ii) {
+          mf_tmp <- matrix_form(obj[ii, ], TRUE, TRUE, indent_size = indent_size)
+          mf_col_widths(mf_tmp) <- colwidths
```
Melkiades, line 1001: @shajoezhu very important fix that was avoiding the wrapping when paginated

R/tostring.R
```diff
@@ -724,50 +724,51 @@ wrap_string <- function(str, width, collapse = NULL, smart = TRUE) {
...
+        we_interval <- we_interval[1]
+      }
+      # Paste together and rerun if it is not the same as original ret
+      ret_collapse <- paste0(ret, collapse = " ")
+
+      # Checking if we are stuck in a loop
+      ori_wrapped_txt_v <- .go_stri_wrap(str, width)
+      cur_wrapped_txt_v <- .go_stri_wrap(ret_collapse, width)
+      if (!setequal(ori_wrapped_txt_v, cur_wrapped_txt_v)) {
+        return(wrap_string(str = ret_collapse, width = width, collapse = collapse))
```
Melkiades, line 765: This avoids infinite loop

R/matrix_form.R
```diff
@@ -15,24 +15,36 @@ mform_handle_newlines <- function(matform) {
...
     how_many_nl <- length(tl_to_add_back)
   }
+
+  # pre-proc in case of wrapping and \n
+  line_grouping <- mf_lgrouping(matform)
+  strmat <- .compress_mat(strmat, line_grouping, "nl")
+  frmmat <- .compress_mat(frmmat, line_grouping, "unique") # never not unique
+  spamat <- .compress_mat(spamat, line_grouping, "unique")
+  alimat <- .compress_mat(alimat, line_grouping, "unique")
+  line_grouping <- unique(line_grouping)
```
Melkiades, line 41: this compression is needed when columns are wrapped. Without it we get a weird bottom alignment if there are \n preceding this

R/matrix_form.R
```diff
@@ -15,24 +15,36 @@ mform_handle_newlines <- function(matform) {
...
+  nl_inds_header <- seq(1, mf_nlheader(matform))
   hdr_inds <- 1:nr_header
 
   # hack that is necessary only if bottom aligned
   topleft_has_nl_char <- FALSE
   if (has_topleft) {
-    tl <- strmat[hdr_inds, 1, drop = TRUE]
-    strmat[hdr_inds, 1] <- ""
+    tl <- strmat[nl_inds_header, 1, drop = TRUE]
+    strmat[nl_inds_header, 1] <- ""
```
Melkiades, line 28: This fix is needed when the number of rows are not the number of lines of the header (we need to get topleft out and add it back later to respect the right line grouping)

R/matrix_form.R
```diff
@@ -73,15 +85,37 @@ mform_handle_newlines <- function(matform) {
...
+        paste0(col_vec, collapse = "\n")
+      } else {
+        val <- unique(col_vec)
+        val <- val[nzchar(val)]
+        if (length(val) > 1) {
+          stop("Problem in linegroupings! Some do not have the same values.") # nocov
+        } else if(length(val) < 1) {
+          val <- "" # Case in which it is only ""
+        }
+        val[[1]]
```
Melkiades, line 112: These are done for formats (that can be not unique, e.g. `c("1", "")` and `val[[1]]` is needed when there are functions (automatic cast to lists)

R/tostring.R
```diff
@@ -557,6 +557,12 @@ setMethod("toString", "MatrixPrintForm", function(x,
 
     # retrieving titles and footers
     allts <- all_titles(x)
+
+    # Fix for ref_fnotes with \n characters XXX this does not count in the pagination
+    if (any(grepl("\n", ref_fnotes))) {
+      ref_fnotes <- unlist(strsplit(ref_fnotes, "\n", fixed = TRUE))
+    }
+
```
Melkiades, line 565: this is a safenet, being it in toString, it will not update the number of lines for pagination but neither break (I do not expect this to be used at all in general)

R/tostring.R
```diff
@@ -678,11 +684,6 @@ new_line_warning <- function(str_v) {
 #' @param collapse character(1) or `NULL`. If the words that have been split should
 #'   be pasted together with the collapse character. This is usually done internally
 #'   with `"\n"` to have the wrapping updated along with other internal values.
-#' @param smart logical(1). Defaults to `TRUE`. It attempts to calculate the optimal
```
Melkiades, line 681: we use only the "smart" recursive algorithm

R/tostring.R
```diff
@@ -778,14 +779,40 @@ wrap_string <- function(str, width, collapse = NULL, smart = TRUE) {
   return(ret)
 }
 
+.go_stri_wrap <- function(str, w) {
```
Melkiades, line 782: compacted it

R/tostring.R
```diff
@@ -724,50 +724,51 @@ wrap_string <- function(str, width, collapse = NULL, smart = TRUE) {
...
+      # Checking if we are stuck in a loop
+      ori_wrapped_txt_v <- .go_stri_wrap(str, width)
+      cur_wrapped_txt_v <- .go_stri_wrap(ret_collapse, width)
+      if (!setequal(ori_wrapped_txt_v, cur_wrapped_txt_v)) {
+        return(wrap_string(str = ret_collapse, width = width, collapse = collapse))
+      } else {
+        # help function: Very rare case where the recursion is stuck in a loop
+        ret_tmp <- force_split_words_by(ret[we_interval], width) # here we_interval is only one ind
+        ret <- append(ret, ret_tmp, we_interval)[-we_interval]
+        which_exceeded <- which(nchar(ret) > width)
```
Melkiades, line 770: force_split was necessary because `stringi` appears unable to split certain characters such as "x ." into smaller elements. I suspect that space and full stop are considered one word (even if normalize = TRUE) so it gets stuck in an eternal recursion, so we need to break it manually, append it and redefine the index of what is needed to be modified

R/tostring.R
```diff
@@ -778,14 +779,40 @@ wrap_string <- function(str, width, collapse = NULL, smart = TRUE) {
...
+force_split_words_by <- function(ret, width) {
+  which_exceeded <- which(nchar(ret) > width)
+  ret_tmp <- NULL
+  for (ii in seq_along(ret)) {
+    if (ii %in% which_exceeded) {
+      wrd_i <- ret[ii]
+      init_v <- seq(1, nchar(wrd_i), by = width)
+      end_v <- c(init_v[-1] - 1, nchar(wrd_i))
+      str_v_tmp <- stringi::stri_sub(wrd_i, from = init_v, to = end_v)
+      ret_tmp <- c(ret_tmp, str_v_tmp[!grepl("^\\s+$", str_v_tmp) & nzchar(str_v_tmp)])
```
Melkiades, line 801: avoiding empties is needed at this stage to avoid spurious spaces appearing (it should not happen too often to get here anyway, only with seriously small widths)

tests/testthat/test-exporters.R
```diff
@@ -98,13 +98,14 @@ test_that("exporters work", {
...
                                 line_grouping = c(1, 1, 2, 3, 4),
                                 nrow_header = 2,
                                 row_info = rinfo)
 
         mymf_out <- toString(mymf, hsep = "-")
-        expct_lns <- c("  hi          lo     ",
-                       "            there    ",
+        expct_lns <- c("              lo     ",
+                       "  hi        there    ",
                        "(N=50)   (N=whoknows)",
```
Melkiades, line 109: I value this as an improvement

tests/testthat/test-exporters.R
```diff
@@ -98,13 +98,14 @@ test_that("exporters work", {
         mymf <- MatrixPrintForm(strings = strs, aligns = matrix("center", ncol = 2, nrow = 5),
                                 formats = matrix("xx", ncol = 2, nrow = 5),
                                 spans = matrix(1L, ncol = 2, nrow = 5),
+                                has_topleft = FALSE,
                                 line_grouping = c(1, 1, 2, 3, 4),
```
Melkiades, line 102: not having to add topleft info was a problem as it was mistakingly considering the first colname as a topleft info

vignettes/formatters.Rmd
```diff
@@ -135,7 +135,7 @@ matrix_form.data.frame <- function(df) {
     strings = strings,
     aligns = aligns,
     spans = matrix(1, nrow = fnr, ncol = fnc),
-    formats = NULL,
+    formats = matrix("", nrow = fnr, ncol = fnc),
```
Melkiades, line 138: This should not be allowed in general

shajoezhu (approved):
thanks a lot @Melkiades , i am merging this in to trigger UAT

PR #225, opened 2023-11-09
Support for combining duplicate ref footnotes

edelarua:
Part of https://github.com/insightsengineering/rtables/issues/779

### DIFF
--- a/R/tostring.R
+++ b/R/tostring.R
@@ -671,15 +671,18 @@ reorder_ref_fnotes <- function(fns) {
   ind <- gsub("\\{(.*)\\}.*", "\\1", fns)
   ind_num <- suppressWarnings(as.numeric(ind))
   is_num <- !is.na(ind_num)
+  is_asis <- ind == fns
 
   if (all(is_num)) {
     ord_num <- order(ind_num)
     ord_char <- NULL
+    ord_other <- NULL
   } else {
     ord_num <- order(ind_num[is_num])
-    ord_char <- order(ind[!is_num])
+    ord_char <- order(ind[!is_num & !is_asis])
+    ord_other <- order(ind[is_asis])
   }
-  c(fns[is_num][ord_num], fns[!is_num][ord_char])
+  c(fns[is_num][ord_num], fns[!is_num & !is_asis][ord_char], ind[is_asis][ord_other])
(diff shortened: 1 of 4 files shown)


R/tostring.R
```diff
@@ -571,6 +571,7 @@ setMethod("toString", "MatrixPrintForm", function(x,
   # retrieving titles and footers
   allts <- all_titles(x)
 
+  ref_fnotes <- unique(ref_fnotes)
```
Melkiades, line 574: did you check if this is not messing up pagination indices?

Melkiades: Could you add a test for this? It is a bit buggy already be aware

edelarua: Doesn't seem to change the previous ref footnote behaviour with exporters/pagination - I have added another test in rtables just in case.

edelarua: Doing some more testing now

R/tostring.R
```diff
@@ -666,6 +667,12 @@ setMethod("toString", "MatrixPrintForm", function(x,
   footer_txt
 }
 
+reorder_ref_fnotes <- function(fns) {
+  ind <- gsub("\\{(.*)\\}.*", "\\1", fns)
+  ind_num <- suppressWarnings(as.numeric(ind))
+  if (!any(is.na(ind_num))) fns[order(ind_num)] else fns
```
Melkiades, line 673: it is na when there are * right? If that is the case do we want to order them anyway? Like putting them first ordered as it was and then the others?

R/tostring.R
```diff
@@ -666,6 +667,12 @@ setMethod("toString", "MatrixPrintForm", function(x,
   footer_txt
 }
 
+reorder_ref_fnotes <- function(fns) {
+  ind <- gsub("\\{(.*)\\}.*", "\\1", fns)
+  ind_num <- suppressWarnings(as.numeric(ind))
```
Melkiades, line 672: I think we can better manage the exception here where it is not numeric. I think we can add a numeric detector before

Melkiades (approved):
lgtm! Thanks Emily! I am wondering if you tested how this cooperates with `*` refnotes. Could you add a small test there?

PR #226, opened 2023-11-14
allow content row in the end

### LINKED ISSUE #233: rtables summarize_row_groups fails with export_as_txt
```R
basic_table() %>%
  split_rows_by("Species") %>%
  summarize_row_groups("Sepal.Length") %>%
  build_table(iris) %>%
  export_as_txt()
```

clarkliming:
close #233 

allow content row in the end.

for each row, if it is content row and it do not have any children, then allow it to be paginated at that position.


Melkiades (changes requested):
Thanks Liming for this. I think it makes sense, and this should happen already in theory. Can you provide a couple of tests and examples for this?

tests/testthat/test-pagination.R
```diff
@@ -64,14 +64,13 @@ test_that("pagination works", {
   ##                                min_siblings = 0,
   ##                                verbose = TRUE))
 
-  expect_error(
+  expect_message(
```
clarkliming, line 67: there is one test here. do you want more? @Melkiades

Melkiades: This does not test how it looks for all the cases (e.g. w/ different combinations of content, data and label rows), if the pagination counts are correct and if the exporters work accordingly for these cases. In general, my asking for visual tests, in particular, serves us to review precisely what is the purpose of this modification and to appreciate the extent of changes. For example, I am quite sure that we would need to change a lot of documentation to reflect this change. This needs to be reviewed and added to the PR

edelarua (approved):
This makes sense to me to allow. Looks good!

Melkiades (approved):

PR #227, opened 2023-11-21
add page number

### LINKED ISSUE #160: page number implementation
migrated issue from https://github.com/insightsengineering/rlistings/issues/80

clarkliming:
add page number in tables fixes #160


Melkiades (changes requested):
Thanks Liming for working on the page numbering. We need to have a more general solution here in my opinion. Also, I would like some consensus on how this page numbering should appear before committing to it. What do you think @edelarua @ayogasekaram @shajoezhu?

ayogasekaram:
> Thanks Liming for working on the page numbering. We need to have a more general solution here in my opinion. Also, I would like some consensus on how this page numbering should appear before committing to it. What do you think @edelarua @ayogasekaram @shajoezhu?

I think page x/n is clear so I am comfortable with this solution. I think anything longer or wordier would interfere with the table aesthetics.

R/mpf_exporters.R
```diff
@@ -21,6 +21,8 @@
 #' @param paginate logical(1). Whether pagination should be performed,
 #'     defaults to \code{TRUE} if page size is specified (including
 #'     the default).
+#' @param ...  Further parameters to be passed to [paginate_to_mpfs()].
```
Melkiades, line 24: Should we pass more things? all the pagination parameters are directly passed down and I think they are double documented

R/pagination.R
```diff
@@ -1068,6 +1075,7 @@ paginate_to_mpfs <- function(obj,
                              max_width = NULL,
                              indent_size = 2,
                              pg_size_spec = NULL,
+                             page_num = NULL,
```
Melkiades, line 1078: I do not think page_num should be on as default

R/pagination.R
```diff
@@ -1179,7 +1198,31 @@ paginate_to_mpfs <- function(obj,
     })
   })
 
-  unlist(res, recursive = FALSE)
+  res <- unlist(res, recursive = FALSE)
+
+  # Adding page numbers if needed
+  if (!is.null(page_num)) {
+    total_pages <- length(res)
+    page_str <- gsub("\\{n\\}", total_pages, page_num)
```
Melkiades, line 1206: @clarkliming this is basically your code with some tweaks. Also now it is inside the pagination machinery, so it will be available (if we send down `...`) also in other exports

tests/testthat/test-pagination.R
```diff
@@ -364,3 +364,52 @@ test_that("spans and string matrix match after pagination when table has single
...
+    pg_tst_off <- paginate_to_mpfs(tst,
+      cpp = 19 + 5 + 3, # rownames + max colwidths + 3 (extra colgap)
+      lpp = 20,
+      page_num = "This is too long, it is breaking"
+    ),
+    "Page numbering string (page_num) is too wide to fit the desired page (inserted cpp)."
+  )
+
+  # Very stringent test with export_as_txt
+  pg_tst_exp <- export_as_txt(tst,
```
Melkiades, line 408: Still to add some test for `export_as_pdf` and other exports

Melkiades:
fyi this should be ready @shajoezhu @BFalquet

shajoezhu (approved):
approve this to unblock merging

BFalquet:
It looks good to me, in the downstream packages we would probably have to think a bit how the user will control that but I guess that a first approximation would be to adapt the export function to fetch the value of an option as you suggested. We have already adopted this strategy for the representation of `NA` values and we are moving toward that for other parameters so this seems coherent.

R/mpf_exporters.R
```diff
@@ -56,7 +59,8 @@ export_as_txt <- function(x,
                           nosplitin = character(),
                           rep_cols = num_rep_cols(x),
                           verbose = FALSE,
-                          page_break = "\\s\\n") {
+                          page_break = "\\s\\n",
+                          page_num = "page {i}/{n}") {
```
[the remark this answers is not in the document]
Melkiades: We can if we make ... in export_as_pdf

R/mpf_exporters.R
Melkiades: No they are not double documented. The main docs are in `paginate_indices`. We might want to avoid the import of string parameters and be sure about the `...` use

Melkiades (approved):

Issue #141, opened 2023-04-20
Better documentation for function insertion in the format value (e.g. `sprintf` cooperation)


Melkiades:
I think this is one of the best features in customizing rtables values. We should allow users to know this more readily. I propose to extend examples in function `sprintf_format` and add `@seealso` to `list_valid_format*`  with that function and `round_fmt`. It is also necessary imo to have a short explanation at the end of function call `list_valid_format*` that explains the possibility to use dynamic function calls to set the format.

Tracked by https://github.com/insightsengineering/rtables/issues/611

PR #269, opened 2024-03-15
Extend examples and see also of sprintf_format function

martincadek:
Closes https://github.com/insightsengineering/formatters/issues/141

Related to https://github.com/insightsengineering/rtables/issues/611


martincadek:
Check fails at library load which is required for the example.
`> library(rtables)`
`Error in library(rtables) : there is no package called ‘rtables’`

I believe the formatters won't import the package but I'll do further tests. If I am not able to use `rtables` I will use ` \dontrun{}` wrapper. 

In addition, this part of the issue is somewhat unclear to me '_It is also necessary imo to have a short explanation at the end of function call list_valid_format* that explains the possibility to use dynamic function calls to set the format._'. The function `list_valid_format_labels` states "Currently valid format labels can not be added dynamically." Can you please provide what would you suggest adding @Melkiades. Thank you!

Melkiades:
> Check fails at library load which is required for the example. `> library(rtables)` `Error in library(rtables) : there is no package called ‘rtables’`
> 
> I believe the formatters won't import the package but I'll do further tests. If I am not able to use `rtables` I will use ` \dontrun{}` wrapper.

Do not use rtables from formatters as it would result in circular dependencies

> 
> In addition, this part of the issue is somewhat unclear to me '_It is also necessary imo to have a short explanation at the end of function call list_valid_format* that explains the possibility to use dynamic function calls to set the format._'. The function `list_valid_format_labels` states "Currently valid format labels can not be added dynamically." Can you please provide what would you suggest adding @Melkiades. Thank you!

This issue needs to have documentation for the various sprintf and round formats that we added or can add when format is a function. Indeed, you can use this to change everything that is printed out and we lack documentation for this powerful feature

R/format_value.R
```diff
@@ -120,13 +120,36 @@ check_aligns <- function(algn) {
 #' @return A formatting function which wraps and will apply the specified \code{printf} style format
 #'   string \code{format}.
 #' @seealso \code{\link[base]{sprintf}}
+#' 
+#'  \code{\link[base]{round_fmt}}
```
Melkiades, line 124: we use only `[round_fmt()]`. It comes from formatters too, not base

R/format_value.R
```diff
@@ -120,13 +120,36 @@ check_aligns <- function(algn) {
...
 #'
 #' fmtfun2 <- sprintf_format("%.4f - %.2f")
 #' format_value(list(12.23456, 2.724))
+#' 
+#' # You can also customise rtables values
+#' library(rtables)
+#' fmtfun3 <- sprintf_format("this range => %.1f")
+#' format_value(100, format = fmtfun3)
+#' 
+#' fmtfun4 <- sprintf_format("sadbasd asa a %.2f asda a %.1f")
```
Melkiades, line 140: This is kind of what I meant yes!

Melkiades (commented):
We need to add information about the various possible formats in `?list_valid_format_labels()` and in {rtables} with a looot of examples. If you do `?rtables::analyze` you will find the format parameter defined as such:


> format | FormatSpec. Format associated with this split. Formats can be declared via strings ("xx.x") or function. In cases such as analyze calls, they can character vectors or lists of functions.

But as you can see there are no links to other docs or functions that lists useful formats. We should enhance the network of `@seealso` by adding these info in a common place (I would opt for `?list_formats`) that is linked from many {rtables} functions (the one that have format as argument ideally, while list_formats keeps connection to other formatting functions. Final examples (not to have circular deps) should live in a dedicated vignette in {rtables} along with the format handling/precedence

R/format_value.R
```diff
@@ -120,13 +120,36 @@ check_aligns <- function(algn) {
...
+#' format_value(list(12.23456, 2.724), format = fmtfun4)
+#' 
+#' lyt <- basic_table() %>%
+#'   split_cols_by("Species") %>%
+#'   analyze("Sepal.Length", afun = function(x) {
+#'     list(
+#'       "mean (sd)" = rcell(c(mean(x), sd(x)), format = fmtfun4),
+#'       "range" = rcell(diff(range(x)), format = fmtfun3)
+#'     )
+#'   })
```
Melkiades, line 150: unfortnately you cannot call {rtables} from {formatters}

Melkiades:
@martincadek remember to run the styler from rstudio, so you avoid having annoying lintr issues ;)

martincadek:
Thank you for the reviews so far - working on this to have it resolved today.

martincadek:
I have read the CLA Document and I hereby sign the CLA

R/format_value.R
```diff
@@ -121,12 +121,36 @@ check_aligns <- function(algn) {
 #'   string \code{format}.
 #' @seealso \code{\link[base]{sprintf}}
 #'
+#'  \code{\link[formatters]{round_fmt}}
+#'
+#'  \code{\link[formatters]{list_formats}}
+#'
```
martincadek, line 127: Corrected to link to formatters.

R/format_value.R
martincadek: Used `\donotrun{}`

R/format_value.R
Melkiades: use the above not \link

R/format_value.R
Melkiades: donotrun is not enough, we need to take it out

martincadek: @Melkiades That would only leave the `format_value` in place is that what you had in mind? That is, the `lyt` table is not needed for full example? If not, I'll remove 143 to 153, please let me know.

R/format_value.R
martincadek: I've edited this, thank you for the feedback, pls lmk if it needs any further amendment.

R/format_value.R
martincadek: Lines completely removed.

R/format_value.R
```diff
@@ -119,14 +119,19 @@ check_aligns <- function(algn) {
 #' @export
 #' @return A formatting function which wraps and will apply the specified \code{printf} style format
 #'   string \code{format}.
-#' @seealso \code{\link[base]{sprintf}}
-#'
+#' @seealso \code{\link[base]{sprintf}} [round_fmt()] \code{link{list_formats}}
```
Melkiades, line 122: @martincadek could you make it all [fnc()] please? Also can you go in the other functions and see that they have some good connection with this one? otherwise it remains difficult to reach

Melkiades: I think there should also be a counterpart in {rtables} with links to these in some of the obscure description of parameters. All of this would make a nice extension of the current vignette on format precedence

Melkiades:
Closing this as outdated

PR #270, opened 2024-03-18
261 truetype pag new

gmbecker:
Adds the concept of font-based width (number of spaces in the font setting) to wrapping and pagination logic. Lays foundation for upcoming PRs in rtables and rlistings repos.


gmbecker:
The spell check results are purely false-positives from what I can see

Melkiades:
> The spell check results are purely false-positives from what I can see

Code should be between ` and truetype should be one (does it have hypen?) the rest may be added to the spelling
```
df           MatrixPrintForm.Rd:60
listOrNULL   MatrixPrintForm.Rd:62
monospace    export_as_pdf.Rd:119
             nchar_ttype.Rd:19,39
             tostring.Rd:53
truetype     export_as_pdf.Rd:119
             nchar_ttype.Rd:41,44
             tostring.Rd:53
Truetype     wrap_string_ttype.Rd:6,46
unrounded    nchar_ttype.Rd:[26](https://github.com/insightsengineering/formatters/actions/runs/8333697785/job/22815864386?pr=270#step:9:29)
```

Melkiades:
> **CLA Assistant Lite bot:** 🎉 Thank you for your contribution! Before this PR can be accepted, we require that you all read and agree to our [Contributor License Agreement](https://github.com/insightsengineering/.github/blob/main/CLA.md). You can digitally sign the CLA by posting a comment on this Pull Request in the format shown below. This agreement will apply to this PR _as well as_ all future contributions on this repository.
> 
> I have read the CLA Document and I hereby sign the CLA
> 
> **0** out of **2** committers have signed the CLA.❌ @gmbecker❌ @MelkiadesYou can retrigger this bot by commenting **recheck** in this Pull Request

@cicdguy @shajoezhu how can we sign the CLA?

Melkiades:
I have read the CLA Document and I hereby sign the CLA

Melkiades:
recheck

Melkiades:
@gmbecker could you also fix the lintr warnings please? ^^
https://github.com/insightsengineering/formatters/actions/runs/8356002418/job/22872204679?pr=270 
I think it is only a matter of code style (like extra spacing etc)

R/mpf_exporters.R
```diff
@@ -426,20 +437,21 @@ export_as_rtf <- function(x,
...
-  fullmf <- matrix_form(x, indent_rownames = TRUE, fontspec = fontspec)
-  req_ncols <- ncol(fullmf) + as.numeric(mf_has_rlabels(fullmf))
-  if (!is.null(colwidths) && length(colwidths) != req_ncols) {
-    stop(
-      "non-null colwidths argument must have length ncol(x) (+ 1 if row labels are present) [",
-      req_ncols, "], got length ", length(colwidths)
-    )
-  }
+  # NEEDS TO BE INTO paginate_to_mpfs so to have this check once for all paginations
+  # fullmf <- matrix_form(x, indent_rownames = TRUE, fontspec = fontspec)
```
Melkiades, line 441: we have a check inside `paginate_to_mpfs` that does this check so for all exports

R/mpf_exporters.R
```diff
@@ -543,8 +555,9 @@ export_as_pdf <- function(x,
                           rep_cols = NULL,
                           tf_wrap = TRUE,
                           max_width = NULL,
+                          colwidths = NULL,
                           fontspec = font_spec(font_family, font_size, lineheight),
-                          colwidths = propose_column_widths(x, indent_size = indent_size, fontspec = fontspec),
+                          #colwidths = propose_column_widths(x, indent_size = indent_size, fontspec = fontspec),
```
Melkiades, line 560: This happens in `paginate_to_mpfs` instead of here

R/generics.R
```diff
@@ -651,6 +675,28 @@ setMethod("num_rep_cols", "MatrixPrintForm", function(obj) {
...
+#' @rdname num_rep_cols
+setGeneric("num_rep_cols<-", function(obj, value) standardGeneric("num_rep_cols<-"))
+#' @export
+#' @rdname num_rep_cols
+setMethod("num_rep_cols<-", "ANY", function(obj, value) stop("No num_rep_cols<- method for class ", class(obj)))
+
+#' @export
+#' @rdname num_rep_cols
+setMethod("num_rep_cols<-", "MatrixPrintForm", function(obj, value) {
+    obj <- mf_update_cinfo(obj, colwidths = NULL, rep_cols = value)
```
Melkiades, line 694:
```suggestion
  obj <- mf_update_cinfo(obj, colwidths = NULL, rep_cols = value)
```

R/pagination.R
```diff
@@ -1082,8 +1148,16 @@ paginate_to_mpfs <- function(obj,
                              pg_size_spec = NULL,
                              page_num = default_page_number(),
                              rep_cols = NULL,
+                             # rep_cols = num_rep_cols(obj),
+                             # col_gap = 3, # this could be change in default - breaking change
                              col_gap = 2,
```
Melkiades, line 1153: What is the standard? Is it 3? We should align all defaults to 3 or 2

R/generics.R
```diff
@@ -651,6 +675,28 @@ setMethod("num_rep_cols", "MatrixPrintForm", function(obj) {
   }
 })
 
+#' @export
+#' @rdname num_rep_cols
+setMethod("num_rep_cols", "MatrixPrintForm", function(obj) obj$num_rep_cols)
```
Melkiades, line 680: This needs to be merged with the above. Or this new value needs to be assigned with keycols in matrix_form from {rlistings}. If the second we need to change `.get_keycols_from_listing`

Melkiades: We still need which keycols are there so I would keep the new entry in matrix_form. This would be probably a one-liner in matrix_form in rlistings (for num_rep_cols)

R/tostring.R
```diff
@@ -826,14 +1074,144 @@ wrap_string <- function(str, width, collapse = NULL) {
 }
 
 .go_stri_wrap <- function(str, w) {
+    if(is.na(w) || is.nan(w))
+        browser()
```
Melkiades, line 1078: I think it may be forgotten here

R/generics.R
```diff
@@ -81,7 +86,9 @@ setMethod("make_row_df", "MatrixPrintForm", function(tt, colwidths = NULL, visib
                                                      repr_inds = integer(),
                                                      sibpos = NA_integer_,
                                                      nsibs = NA_integer_,
-                                                     max_width = NULL) {
+                                                     max_width = NULL,
+                                                     fontspec = font_spec(),
+                                                     col_gap = mf_colgap(tt) %||% 3L) {
```
Melkiades, line 91: I think we need a general rule for this. Deciding where it should be set and where it is only "transmitted" and if we keep it from object getters or not. At the moment I find it very hard to track the origin of a default value. I would suggest keeping a standard default (like NULL) until there are no more user-facing functions where if it is not set yet it is set from the object or a default

R/tostring.R
```diff
@@ -60,22 +275,25 @@ do_cell_fnotes_wrap <- function(mat, widths, max_width, tf_wrap) {
...
     ncol = ncol(mfs)
   )
 
-  ## XXXXX this is wrong and will break for listings cause we don't know when
-  ## we need has_topleft to be FALSE!!!!!!!!!!
-  mat <- mform_handle_newlines(mat)
+  if(expand_newlines) {
+    ## XXXXX this is wrong and will break for listings cause we don't know when
+    ## we need has_topleft to be FALSE!!!!!!!!!!
+    mat <- mform_handle_newlines(mat)
```
Melkiades, line 287: I think it is set as FALSE right?

R/tostring.R
```diff
@@ -329,6 +547,12 @@ decimal_align <- function(string_mat, align_mat) {
   string_mat
 }
 
+## this gives the conversion from number of spaces to number of characters
+## for use in, e.g., repping out divider lines.
+calc_str_adj <- function(str, fontspec) {
+    nchar(str)/nchar_ttype(str, fontspec, raw = TRUE)
```
Melkiades, line 553:
```suggestion
  nchar(str) / nchar_ttype(str, fontspec, raw = TRUE)
```

R/tostring.R
```diff
@@ -826,14 +1074,144 @@ wrap_string <- function(str, width, collapse = NULL) {
...
+    list(ok = substr(str, 1, length(ok)),
+         remainder = substr(str, length(ok) + 1, nchar(str)))
+}
+
+## need a separate path here because **the number of characters**
+## in each part is no longer going to be constant the way it
+## was for monospace
+## this is much slower but still shouldn't be a bottleneck, if it is we'll
+## have to do something else
+#' wrap string given a Truetype font
```
Melkiades, line 1109: Could you add here a description of roughly how it works? I think it is important also for me to understand how it works ;)

edelarua:
Documentation/styling/lint is now fixed.

FYI @Melkiades this PR causes breaking changes in `rtables` that will need to be addressed.

R/generics.R
gmbecker: I don't really understand this. It seems like keycolumns and rep_cols are mechanically the same, its just that one is specific to rlistings terminology (key column) while the other is general for anything that plugs into formatters.

The change here that I'm making afair is that instead of just automatically being calculated at toString/pagination time the number of rep-cols is carried around in the MPF now. Maybe I'm missing something else.

R/pagination.R
gmbecker: Thats what I'm saying here, the defaults didn't match around the codebase. 2 seemed more common than 3, so I changed this 3 to 2.

I agree we should make sure the default is the same everywhere.

R/tostring.R
gmbecker: Looks like this is outdated?

R/generics.R
gmbecker: I think it should be able to be set in any user-facing call, and should override the one carried around if that is the case. 

We could get rid of the `%||% 3L` part here if we want logic in the body that handles when `length(mf_colgap(tt)) == 0`, though honestly I kinda like this better since you can see what it does in the default value, whereas the othr way you have to look in the body of the function...

R/tostring.R
gmbecker: Yeah this issue has been solved for a long time, I think. We have accessors that can tell us whether we have topleft material (in practice)

R/pagination.R
Melkiades: This comment is coming from my merge, I think you changed this default to 3 in this PR. For me it is the same ;)

R/generics.R
Melkiades: I think it makes sense in this case. What do we add in (for example) export_txt ? We cannot reference the object tt because now it can be a list of tables/listings, so we should have either NULL and then we add what you added here in a `if(is.null)` inside `paginate_to_mpfs` or just always 3L/2L. I am leaning towards the second for this case, but then I am a bit confused whether we assign/retrieve from the object itself at all and how this depends on the creation of the object table/listing. We would need a graph of these dependencies. Many would respect the same pipeline I suspect

gmbecker:
I have read the CLA Document and I hereby sign the CLA

gmbecker:
> Documentation/styling/lint is now fixed.
> 
> FYI @Melkiades this PR causes breaking changes in `rtables` that will need to be addressed.

This should only be the case until the rtables PR is also in place, but it looks like that never got filed, or was deleted, I'm not sure. The relevant branch exists in my fork, though, either way

Melkiades:
> > Documentation/styling/lint is now fixed.
> > FYI @Melkiades this PR causes breaking changes in `rtables` that will need to be addressed.
> 
> This should only be the case until the rtables PR is also in place, but it looks like that never got filed, or was deleted, I'm not sure. The relevant branch exists in my fork, though, either way

it is not this, right? https://github.com/insightsengineering/rtables/pull/819 
I cannot see any connected PR to this one (also not between the closed ones). If it is between your PRs in the fork, do not hesitate to send the MR ;)

gmbecker:
There isn't a PR for it, though you can a state of it here: https://github.com/gmbecker/rtables/tree/261_truetype_pag

The issue, as you can see, is that there is quite a bit of divergence caused by the doc stuff that I haven't had time to resolve. I'm going to look at it some today and see what I can get working

gmbecker:
The corresponding rtables PR now exists: insightsengineering/rtables#865

I will create the one for rlistings shortly

R/generics.R
Melkiades: It is probably going to work either way because num_rep_cols is assigned to the number of keycolumns. I think this may not be valid in rlistings hence my addition here, but I need to double-check. Either way is fine, we need to merge the two functions nonetheless as they work on the same object

Melkiades: after this fix https://github.com/insightsengineering/rlistings/pull/211#discussion_r1600357912 we can remove the method strictly above this one

gmbecker:
check failure should be resolved by latest commit

R/generics.R
gmbecker: So MPFs carry around their col_gaps after creation, and if it has one, I think that should be used, so I don't think a default of 3/2 taht overrides taht would be desirable

R/tostring.R
```diff
@@ -1,3 +1,247 @@
...
+#' In both cases the value is returned invisibly.
+#'
+#' @examples
+#' open_font_dev(font_spec("Times"))
+#' nchar_ttype("Hiya there", font_spec("Times"))
+#' close_font_dev()
+#'
+#' @export
+open_font_dev <- function(fontspec, silent = FALSE) {
+  ## XXX to remove for debugging only
```
Melkiades, line 57: should we keep it only for debugging or remove it? if only debugging I would keep internal, It is always accessible with ::: or by doing devtools::load_all()

gmbecker: Sorry, my comment may hae been unclear, `open_font_dev` is an important exported api entrypoint, the comment is about the lines directly below it, which yes can probably be removed, the real "For maintainers debuggin switch" is at line 77 where you remove the FALSE. Will push a commit to remove that comment

gmbecker: pushed

Melkiades: understood!!! thanks

R/tostring.R
```diff
@@ -349,6 +600,9 @@ decimal_align <- function(string_mat, align_mat) {
 #'   footnotes) materials should be word-wrapped to. If `NULL`, it is set to the current print width of the
 #'   session (`getOption("width")`). If set to `"auto"`, the width of the table (plus any table inset) is
 #'   used. Parameter is ignored if `tf_wrap = FALSE`.
+#' @param ttype_ok logical(1). Should truetype (non-monospace) fonts be
+#' allowed via `fontspec`. Defaults to `FALSE`. This parameter is primarily
+#' for internal testing and generally should not be set by end users.
```
Melkiades, line 605:
```suggestion
#' @param ttype_ok (`logical(1)`)\cr should truetype (non-monospace) fonts be
#'   allowed via `fontspec`. Defaults to `FALSE`. This parameter is primarily
#'   for internal testing and generally should not be set by end users.
```

R/tostring.R
```diff
@@ -1008,6 +1452,7 @@ propose_column_widths <- function(x, indent_size = 2) {
             # need to update widths to fit content with colspans
             # spread width among columns
             widths[i_cols] <- cw_i + spread_integer(nchar_i - available_width, length(cw_i))
+            ## widths[i_cols] <- cw_i + spread_width(nchar_i - available_width, length(cw_i))
```
Melkiades, line 1455:
```suggestion
```
Did you mean to comment here?

R/tostring.R
```diff
@@ -1038,16 +1568,16 @@ propose_column_widths <- function(x, indent_size = 2) {
...
   just <- match.arg(just)
 
   if (length(x) != 1) stop("length of x needs to be 1 and not", length(x))
   if (is.na(n) || !is.numeric(n) || n < 0) stop("n needs to be numeric and > 0")
 
   if (is.na(x)) x <- "<NA>"
 
-  nc <- nchar(x)
-
+  ## nc <- nchar(x)
```
Melkiades, line 1579:
```suggestion
```
I think it is fine to remove, It is described as an extension of nchar

R/tostring.R
Melkiades: Because if commented this function (spread_width) does not seem used

R/tostring.R
```diff
@@ -805,14 +1087,159 @@ wrap_string <- function(str, width, collapse = NULL) {
...
+                              fontspec,
+                              collapse = NULL,
+                              min_ok_chars = min(floor(nchar(str) / 2), 4, floor(width / 2)),
+                              wordbreak_ok = TRUE) {
+  newdev <- open_font_dev(fontspec)
+  if (newdev) {
+    on.exit(close_font_dev())
+  }
+
+  rawspls <- strsplit(str, "[[:space:]](?=[^[:space:]])", perl = TRUE)[[1]] # preserve all but one space
```
Melkiades, line 1150: It is of course fine but why do you enforce it? stringi should be able to handle multiple spaces correctly

R/tostring.R
```diff
@@ -805,14 +1087,159 @@ wrap_string <- function(str, width, collapse = NULL) {
...
+  nctt <- nchar_ttype(rawspls, fontspec, raw = TRUE)
+  pts <- which(cumsum(nctt) <= width)
+  if (length(pts) == length(rawspls)) { ## no splitting needed
+    return(str)
+  } else if (length(pts) == 0) { ## no spaces, all one word, split it and keep going
+    if (wordbreak_ok) {
```
Melkiades, line 1156: This feature is not avalilable at the moment externally right? It is "just in case"? (it makes complete sense to have it imo)

R/tostring.R
```diff
@@ -805,14 +1087,159 @@ wrap_string <- function(str, width, collapse = NULL) {
...
+      c(
+        inner_res$remainder,
+        if (length(rawspls) > length(pts) + 1) tail(rawspls, -(length(pts) + 1))
+      ),
+      collapse = " "
+    )
+  }
+  ret <- c(
+    done,
+    wrap_string_ttype(remainder, width, fontspec)
```
Melkiades, line 1196: eheh at the end I also went for a recursive solution. I did not expect it but it is incredibly faster too

R/tostring.R
```diff
@@ -805,14 +1087,159 @@ wrap_string <- function(str, width, collapse = NULL) {
 }
 
 .go_stri_wrap <- function(str, w) {
+  if (is.na(w) || is.nan(w)) {
+    browser()
+  }
```
Melkiades, line 1092: it should not happen right?
```suggestion
```

R/tostring.R
```diff
@@ -805,14 +1087,159 @@ wrap_string <- function(str, width, collapse = NULL) {
...
+  if (w < 1) {
```
Melkiades, line 1093: Can this happen?

R/tostring.R
```diff
@@ -805,14 +1087,159 @@ wrap_string <- function(str, width, collapse = NULL) {
...
+## need a separate path here because **the number of characters**
+## in each part is no longer going to be constant the way it
+## was for monospace
+## this is much slower but still shouldn't be a bottleneck, if it is we'll
+## have to do something else
+#' wrap string given a Truetype font
+#'
+#' @inheritParams wrap_string
+#' @param min_ok_chars numeric(1). Number of minimum characters which much remain
+#' on either side when a word is split
```
Melkiades, line 1133:
```suggestion
#' @param min_ok_chars (`numeric(1)`). Number of minimum characters that remain
#'   on either side when a word is split.
```

R/tostring.R
```diff
@@ -805,14 +1087,159 @@ wrap_string <- function(str, width, collapse = NULL) {
...
+#' @param wordbreak_ok logical(1). Should breaking within a word be allowed. If, `FALSE`,
+#' attempts to wrap a string to a width narrower than its widest word will result
+#' in an error.
```
Melkiades, line 1136:
```suggestion
#' @param wordbreak_ok (`logical(1)`)\cr should breaking within a word be allowed? If, `FALSE`,
#'   attempts to wrap a string to a width narrower than its widest word will result
#'   in an error.
```

R/tostring.R
Melkiades: I read it in detail now. I have some question but it is somehow clear ;) we are meeting anyway for details

R/tostring.R
```diff
@@ -59,22 +300,25 @@ do_cell_fnotes_wrap <- function(mat, widths, max_width, tf_wrap) {
...
     ncol = ncol(mfs)
   )
 
-  ## XXXXX this is wrong and will break for listings cause we don't know when
-  ## we need has_topleft to be FALSE!!!!!!!!!!
-  mat <- mform_handle_newlines(mat)
-
-  ## this updates extents in rinfo AND nlines in ref_fnotes_df
-  mat <- update_mf_nlines(mat, max_width = max_width)
+  if (expand_newlines) {
```
Melkiades, line 309: is there a case where we do not want it?

R/page_size.R
```diff
@@ -3,8 +3,63 @@
 NULL
 ## https://www.ietf.org/rfc/rfc0678.txt
 
-## This assumes fixed font size, monospaced font
+times_font_name <- function() {
+  plat <- .Platform$OS.type
+  switch(plat,
+    "unix" = "Times",
```
Melkiades, line 9: Is mac also here? I think it should be unix-like right?

R/page_size.R
```diff
@@ -72,6 +127,8 @@ page_dim <- function(page_type) {
   pg_dim_names[[page_type]]
 }
 
+ascii_chars <- vapply(33:126, function(n) rawToChar(as.raw(n)), "")
+
```
Melkiades, line 131: Is this intentionally here?

R/page_size.R
```diff
@@ -3,8 +3,63 @@
...
+    list(
+      family = font_family,
+      size = font_size,
+      lineheight = lineheight
+    ),
+    class = c("font_spec", "list")
+  )
+}
 std_cpi <- 10L
 std_lpi <- 6L
```
Melkiades, line 64: There were already some default values. Should we group these somewhere more visible? even only the top of the script with some description of their uses

R/mpf_exporters.R
```diff
@@ -510,7 +538,8 @@ export_as_pdf <- function(x,
                           min_siblings = 2,
                           font_family = "Courier",
                           font_size = 8,
-                          fontsize = lifecycle::deprecated(),
+                          fontsize = font_size,
+                          lineheight = 1.2, ## XXX this matches legacy behavior but differs from default everywhere else
```
Melkiades, line 542: should we have a system getter for this default?

R/mpf_exporters.R
```diff
@@ -388,14 +398,17 @@ mpf_to_rtf <- function(mpf,
...
                           colwidths = NULL,
                           page_type = "letter",
                           pg_width = page_dim(page_type)[if (landscape) 2 else 1],
                           pg_height = page_dim(page_type)[if (landscape) 1 else 2],
                           landscape = FALSE,
                           margins = c(bottom = .5, left = .75, top = .5, right = .75),
-                          font_size = 8,
                           font_family = "Courier",
+                          font_size = 8,
+                          lineheight = 1,
```
Melkiades, line 410: different lineheight?

R/matrix_form.R
```diff
@@ -525,7 +607,7 @@ mf_col_widths <- function(mf) {
       "number of columns in strings matrix (", NCOL(mf_strings(mf)), ")."
     )
   }
-  mf$col_widths <- value
+  mf <- mf_update_cinfo(mf, colwidths = value, rep_cols = NULL)
```
Melkiades, line 610: I need to understand exactly what is the status of cinfo because I have troubles understanding when it is to be used instead of matrix_form entries

R/tostring.R
```diff
@@ -1,3 +1,244 @@
...
+## print width of strings, and we can't be opening
+## a new one every time we want to
+font_dev_state <- new.env()
+font_dev_state$open <- FALSE
+font_dev_state$fontspec <- list()
+font_dev_state$spacewidth <- NA_real_
+font_dev_state$ismonospace <- NA
+font_dev_state$max_ratio <- NA_real_
+font_dev_state$dev_num <- NA_integer_
+
```
Melkiades, line 12: I understand the underlying need for this. I am only wondering if there is an alternative. Maybe @pawelru has an alternative?

R/tostring.R
```diff
@@ -1,3 +1,244 @@
...
+  pushViewport(plotViewport(gp = gp))
+  spcwidth <- cwidth_inches_unsafe(" ")
+  ## XXX this assumes M or W is the widest "reasonable char"
+  ## should be true for any reasonable font but not guaranteed!!!
+  mwidth <- cwidth_inches_unsafe("M")
+  wwidth <- cwidth_inches_unsafe("W")
+  assign("open", TRUE, envir = font_dev_state)
+  assign("fontspec", fontspec, envir = font_dev_state)
+  assign("spacewidth", spcwidth, envir = font_dev_state)
+  assign("ismonospace", spcwidth == cwidth_inches_unsafe("W"),
```
Melkiades, line 106: why not `wwidth`?

R/tostring.R
gmbecker: If this was covered sufficiently in ourmeeting, please resolve it, otherwise we can discuss more on wed

R/matrix_form.R
gmbecker: cinfo (somewhat unfortunate naming on my part, though not worth changing I think) is the column data frame. its used for horizontal pagination of the mpf the same way rinfo is used for vertical pagination.

the pagination data.frame needs to be adjusted when column widths are set/changed, because it changes the self_extent column used in horiz pagination.

R/page_size.R
gmbecker: Yes, mac is also a "unix" os type

R/page_size.R
gmbecker: These are ancient and unused (and also, you'll note, unchanged in the diff)They come from a *very* early initial pass at supporting page sizes. They can be removed, but this PR is probably technically not the right place to do so.

R/tostring.R
gmbecker: Yes, I use it in the jj soon-to-be opensourced rtf exporting function. can go over that more on wed if desired, though its not inside this pr.

R/tostring.R
gmbecker: hmm. well the top case clearly doesn't happen in any of our tests cause the test suite doesn't hang, lol.

the browser can definitely be taken out. whether w can be 0... I feel like in the truetype wrapping algorithm it can, but I'd have to dig in to be sure, lets explore that wed

R/tostring.R
gmbecker: its not available at the `toString` or `paginate_*` level currently, but it can be used when, e.g., iteratively finding optimal column widths (which is massively more non-trivial with truetype fonts) by checking if a particular width *would* break within a word or not given a not-already-wrapped mpf.

It also could be added to `toString`/`paginate_*`in the future if we want

R/tostring.R
gmbecker: strsplit is a base R function, its not from stringi. I wouldn't be surprised if stringi has something similar, but this seemed fine.

In terms of why we care about multiples, we (arguably) want

`"hello   there"` (3 spaces) to wrap to `"hello  ", "there"`  if the width is 7, not  `"hello", "there"`. This is a weird corner case that we could easily say isn't supported at all (ie no putting extra space in labels to fiddle with wrapping behavior) but someday people are gonna want to fiddle with wrapping behavior...

R/tostring.R
gmbecker: You're correct, spread_width spreads the remaining with allowing fractional allocation (ie if there are 3 spaces left and there are 5 columns including row labels, each column gets an extra 3/5 of a space width). 

I was initially intending to allow non-integer column widths, but that feature didn't get into this round of development. Still a nice-to-have in truetype space, as spaces are small but not *that* small, but so far integer colwidths seems to work ok because of how we are exporting rtf (we export an rtf table object, not blocks of text like `export_as_pdf` currently does).

R/tostring.R
gmbecker: thats a logical. its checking (and caching the answer) whether " " and "W" are the same width (if they are its assumed monospace, though I Suppose technically one could artificially construct a font where "W" is the width of " " but other characters are wider or shorter... I dont' think we need to protect ourself against that eventuality though

gmbecker: Oh, nm I misunderstood. 

Yeah, we could use `wwidth` except that I am removing those cause none of that stuff is used (except for by the max ratio calculation which itself is not used anywhere. :)

R/mpf_exporters.R
gmbecker: well, `export_as_rtf` is kinda a mess, but the headline here is that historically the pdf exporter was using the default within grid if lineheight isn't set, which seems to be 1.2, but other places I think the default we set has been 1.

Probably need to unify but that is a potentially breaking change so my PR just leaves the mismatching defaults as they are for lineheight

R/mpf_exporters.R
gmbecker: If you give a fontspec it will override this argument, I guess we could have a "if there is a font device open grab its font_spec" getter. I'm leery of that much state though. i don't even like how much state I *had* to put in already

R/generics.R
gmbecker: It doesn't seem like this can be safely removed, this function (and the corresponding setter) are called in ~9 places in pagination.R...

At least sometimes its called on MatrixPrintForm objects (e.g. default value in `mf_infer_cinfo`)

gmbecker (commented):
github decided some of my responses are part of "a review"...

R/tostring.R
Melkiades: understood! So we keep it for now?

R/tostring.R
Melkiades: but "preserve only one space" seems to indicate that the result is `"hello ", "there"`, hence removing the other spaces. If it keeps them, I would rephrase the comment for clarity. Again it is not super relevant to me, as you wish ^^. For context, we went towards supporting this fiddling at the end by using stringi as people were, as you said, asking to do weird corner cases

R/tostring.R
Melkiades: this makes a lot of sense! propose_colwidths is heavily used and sometimes users add very low columnwidths manually, which can generate issues that are hard to understand (like pagination may break)

R/tostring.R
Melkiades: I think it makes sense to keep it as a safenet, even if it is not a probable scenario

R/page_size.R
Melkiades: it makes sense. I will create a new issue

R/mpf_exporters.R
Melkiades: I understand. We will have to sit down at a certain point and set more general defaults

Melkiades: We had to add a way to set the hsep globally because the continuous hyphen was problematic for non-UTF8 system (dont ask me why they are a thing). Just wondering if it would make sense to have an object with defaults that lives in the package. It is of course a discussion for another time but let me know if you think it could make sense for the future

R/matrix_form.R
Melkiades: I see! I had to fiddle a bit with self_extent for {rlistings} when wrapping keycolumns. You will see in paginate_mpfs something regarding that. Do you think that kind of behavior should rely on cinfo or rinfo? Because at the moment I based it on rinfo

R/tostring.R
gmbecker: its preserving *all but one* space, not only one space. I don't really feel that strongly about how the comment is phrased though

Melkiades: I see, It makes sense now! it is fine like it is ^^

shajoezhu:
can we close this one now?

gmbecker:
@Melkiades if you're satisfied the comments have all been addressed please feel free to close this

PR #277, opened 2024-04-12
Trial fix req validation

Melkiades:
Consistency checks failed:
- requirement #0.17 'Make row layout summary data frames for use during pagination' is not covered by any test

### DIFF
--- a/R/generics.R
+++ b/R/generics.R
@@ -3,7 +3,6 @@
 ### sufficient for hooking that class up to the `formatters` pagination and rendering
 ### machinery.
 
-## nocov start
 #' Make row layout summary data frames for use during pagination
 #'
 #' All relevant information about table rows (e.g. indentations) is summarized in a `data.frame`.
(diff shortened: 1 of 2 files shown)


R/generics.R
```diff
@@ -3,7 +3,6 @@
 ### sufficient for hooking that class up to the `formatters` pagination and rendering
 ### machinery.
 
-## nocov start
```
Melkiades, line 6: I suspect that this was the cause. It is tested and it is covered, but it was not picked up

edelarua (approved):
Lgtm! I don't think this is the issue in `tern` since we only have one section of `nocov`, but we will see if this fixes anything in `formatters`!

R/generics.R
Melkiades: @shajoezhu @ayogasekaram @edelarua this solved the issue for {formatters} about untested documentation (so we can go for releases faster ;)

PR #281, opened 2024-04-29
Add snapshot testing job

cicdguy:
Adds another R CMD check workflow that does non-CRAN checks and therefore executes snapshot tests.


pawelru:
TIL this:
open run details -> scroll down to the bottom
image

This will essentially present the difference of test scope between cran and non-cran type of checks. The expectation is that non-cran would have more or equal number of test cases because of cran skips (explicit ones plus also snapshot tests).

edelarua (approved):
Lgtm, thanks Dinakar!!

PR #287, opened 2024-05-24
Fix recursive page numbering bug

### LINKED ISSUE #286: duplicated page number
pages number are duplicated when object takes more than one page.

```
library(chevron)
library(rtables)
res <- run(lbt01, chevron::syn_data)
withr::with_options(
  list(formatter_default_page_number = "{i}/{n}"),
  cat(export_as_txt(res))
)
```
image

In addition, the `page_num` argument duplicates again the page numbers.

```
withr::with_options(
  list(formatter_default_page_number = "{i}/{n}"),
  cat(export_as_txt(res), page_num = "- page {i} of {n} -")
)
```

image

edelarua:
Closes #286

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,7 @@
 
 * Deprecate `width`, `height` and `fontsize` arguments of `export_as_pdf()` using `lifecycle` package.
 * Add `lifecycle` badge files for documentation.
+* Fixed bug causing recursive appending of page number text during pagination.
 
 ## formatters 0.5.6
  * Added "N=xx" format and unit test for it.
--- a/R/pagination.R
+++ b/R/pagination.R
@@ -1092,7 +1092,7 @@ paginate_to_mpfs <- function(obj,
     return(mpfs)
   }
 
-  if (!is.null(page_num)) {
+  if (!is.null(page_num) && (length(prov_footer(obj)) == 0 || !grepl(page_num, prov_footer(obj), perl = TRUE))) {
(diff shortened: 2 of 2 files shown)


edelarua:
@BFalquet this should fix your issue if you want to test it out. I tested it with the 2 examples you provided and got the correct output. 

Note that for the second example you provided I believe you forgot to put the `page_num` specification within the `export_as_txt` call, which was part of the issue, but the output is correct if this is fixed.

ayogasekaram (commented):
lgtm @edelarua. Could a unit test be added as well?

edelarua:
> Could a unit test be added as well?

A unit test can’t be added in `formatters` since I wasn’t able to replicate the bug without using `rtables`/`chevron` example code, but one could be made elsewhere if needed.

shajoezhu:
> > Could a unit test be added as well?
> 
> A unit test can’t be added in `formatters` since I wasn’t able to replicate the bug without using `rtables`/`chevron` example code, but one could be made elsewhere if needed.

can we please add a test into scda.test, to catch this downstream

Melkiades (approved):
Nice fix Emily! I think the issue comes from `do_force_pagination` which is doing a round of page numbers first and then another in the newer lapply. It was my plan to reconsider the need of that at a certain stage. If you want we can merge this now and open an issue to reassess the need of forced pagination

Melkiades:
For tests, I would ask @BFalquet to add them into {chevron} if he has bandwidth

Issue #288, opened 2024-05-28
Bug in table export


BFalquet:
Reproducible example:
```
library(rtables)
lyt <- basic_table() %>%
  split_cols_by("Species") %>%
  analyze("Sepal.Length", afun = function(x) {
    list(
      "mean (sd)" = rcell(c(mean(x), sd(x)), format = "xx.xx (xx.xx)"),
      "range" = diff(range(x))
    )
  })
lyt

tbl <- build_table(lyt, iris)
prov_footer(tbl) <- c("Santa", "is", "real", "!")
export_as_txt(tbl)
```
leads to

```
Error in length(prov_footer(obj)) == 0 || !grepl(page_num, prov_footer(obj),  : 
  'length = 4' in coercion to 'logical(1)'
```

I think there is just a `any()` missing to handle the case where the prov footer is a vector of several elements.

Melkiades:
@BFalquet I cannot reproduce the error

BFalquet:
@Melkiades  There is also something weird happening with the dependencies today, but just from a purely formal code analysis point of view, the following code in `paginate_to_mpfs` has a small issue:
```
length(prov_footer(obj)) == 0 || !grepl(page_num, prov_footer(obj), perl = TRUE))
```

If the length of prov_footer(obj) is bigger than 1, the `grepl` expression will return a logical vector bigger than 1 that will cause an issue because here the `||` expects two vectors of size 1. E.g
```
> FALSE || grepl("a", letters)
Error in FALSE || grepl("a", letters) : 
  'length = 26' in coercion to 'logical(1)'
```
It doesn't fail every time, because if the first part of the expression is TRUE, the second part is not evaluated but in the other cases, we have an issue.

Anyway, this makes the non-CRAN tests of one of my internal package fail.

Melkiades:
ok I see! let me generalize the fix today

edelarua:
Should be fixed by #290 - bug was still occurring when `prov_footer` is not empty but it's no longer an issue in my tests.

PR #292, opened 2024-05-28
Fix pagination LBT01 issue count

### LINKED ISSUE #291: Pagination issue not showing the right numbers when table is paginated with `do_force_pagination`
We need to understand better this mechanism, but the issue comes from LBT01 pagination inside `length(fpags) > 1`

Melkiades:
now the following produces 6 pages with proper counts:
```r
library(chevron)
library(rtables)
res <- run(lbt01, chevron::syn_data)
options(formatter_default_page_number = "{i}/{n}")
cat(export_as_txt(res))
```

Fixes #291

### DIFF
--- a/R/pagination.R
+++ b/R/pagination.R
@@ -1092,7 +1092,7 @@ paginate_to_mpfs <- function(obj,
     return(mpfs)
   }
 
-  if (!is.null(page_num) && (length(prov_footer(obj)) == 0 || !any(grepl(page_num, prov_footer(obj), fixed = TRUE)))) {
+  if (!is.null(page_num)) {
     # Only adding a line for pagination -> lpp - 1 would have worked too
     prov_footer(obj) <- c(prov_footer(obj), page_num)
   }
@@ -1149,7 +1149,7 @@ paginate_to_mpfs <- function(obj,
       col_gap = col_gap
     )
   }
-  ## this MUST alsways return a list, inluding list(obj) when
+  ## this MUST always return a list, including list(obj) when
   ## no forced pagination is needed! otherwise stuff breaks for things
   ## based on s3 classes that are lists underneath!!!
   fpags <- do_forced_paginate(obj)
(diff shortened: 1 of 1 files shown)


R/pagination.R
```diff
@@ -1161,16 +1161,21 @@ paginate_to_mpfs <- function(obj,
   ## but we will if we ever allow force_paginate to do horiz
   ## pagination.
   if (length(fpags) > 1) {
-    deep_pag <- lapply(
-      fpags, paginate_to_mpfs,
+    # Correction for case we are entering here (page_by?)
+    if (!is.null(page_num)) {
+      prov_footer(obj) <- head(prov_footer(obj), -1)
+      fpags <- do_forced_paginate(obj)
```
Melkiades, line 1167: This is the new fix to #288 and #286

R/pagination.R
```diff
@@ -1161,16 +1161,21 @@ paginate_to_mpfs <- function(obj,
   ## but we will if we ever allow force_paginate to do horiz
   ## pagination.
   if (length(fpags) > 1) {
-    deep_pag <- lapply(
-      fpags, paginate_to_mpfs,
```
Melkiades, line 1165: this was the old list pagination which does not work here. We use directly `paginate_mpfs`

edelarua (approved):
Works great!! Thanks Davide :)

R/pagination.R
gmbecker: It doesn't make sense to call do_forced_paginate(obj), just loop through fpags and reset each of their prov headers...

```
fpags <- lapply(fpags, function(x) {
   prov_footer(x) <- head(prov_footer(x), -1)
   x
})
```

R/pagination.R
```diff
@@ -1161,16 +1161,21 @@ paginate_to_mpfs <- function(obj,
...
+    deep_pag <- paginate_to_mpfs( # what about the other parameters?
+      fpags,
       pg_size_spec = pg_size_spec,
       colwidths = colwidths,
       min_siblings = min_siblings,
       nosplitin = nosplitin,
       verbose = verbose,
       page_num = page_num
     )
-    return(unlist(deep_pag, recursive = FALSE))
```
gmbecker, line 1173: this is a breaking change, previously `paginate_to_mpfs` was guaranteed to return a shallow list of pages, now it can return a nested list

R/pagination.R
Melkiades: You are right, I just did not know exactly how it was working do_forced_paginate so I went extremely conservative eheh. We will have to change all of this during the refactoring we talked about yesterday

PR #300, opened 2024-06-10
Fixes `nlines` function

### LINKED ISSUE #299: `nlines` function fails when width is narrower than the string
**Summary**

There's a problem with `nlines` function initially described here: https://github.com/insightsengineering/rtables/pull/875#issuecomment-2154907466

``` r
formatters::nlines("01234567890", max_width = 5)
#> Error in FUN(X[[i]], ...): argument "fontspec" is missing, with no default
```

<sup>Created on 2024-06-10 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

**R session info**

```sh

# R -e "utils::sessionInfo()" output goes here
R version 4.4.0 (2024-04-24)
Platform: x86_64-pc-linux-gnu
Running under: Ubuntu 24.04 LTS

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.12.0 
LAPACK: /usr/lib/x86_64-linux-gnu/lapack/liblapack.so.3.12.0

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_IE.UTF-8       
 [4] LC_COLLATE=en_US.UTF-8     LC_MONETARY=en_IE.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_IE.UTF-8       LC_NAME=C                  LC_ADDRESS=C              
[10] LC_TELEPHONE=C             LC_MEASUREMENT=en_IE.UTF-8 LC_IDENTIFICATION=C       

time zone: Europe/Berlin
tzcode source: system (glibc)

attached base packages:
[1] stats     graphics  grDevices datasets  utils     methods   base     

loaded via a namespace (and not attached):
 [1] compiler_4.4.0        BiocManager_1.30.23   backports_1.5.0       fastmap_1.2.0        
 [5] cli_3.6.2             htmltools_0.5.8.1     tools_4.4.0           rstudioapi_0.16.0    
 [9] grid_4.4.0            checkmate_2.3.1       formatters_0.5.7.9001 digest_0.6.35        
[13] lifecycle_1.0.4       rlang_1.1.4           renv_1.0.2           
```

**OS / Environment**

- OS: Ubuntu 24.04

averissimo:
# Pull request

- Closes #299 

### Changes description

- Prevent `missing` from being passed in call tree until it fails on `open_font_dev`
- Adds test case

### DIFF
--- a/R/generics.R
+++ b/R/generics.R
@@ -193,7 +193,7 @@ setGeneric(
   "nlines",
   ## XXX TODO come back and add fontspec default value once not having
   ## it has found all the disconnection breakages
-  function(x, colwidths = NULL, max_width = NULL, fontspec, col_gap = NULL) standardGeneric("nlines")
+  function(x, colwidths = NULL, max_width = NULL, fontspec = NULL, col_gap = NULL) standardGeneric("nlines")
 )
 
 ## XXX beware. I think it is dangerous
@@ -202,7 +202,6 @@ setGeneric(
 setMethod(
   "nlines", "list",
   function(x, colwidths, max_width, fontspec, col_gap = NULL) {
-    if (missing(fontspec)) fontspec <- NULL
     if (length(x) == 0) {
       0L
     } else {
@@ -213,12 +212,11 @@ setMethod(
 
 #' @export
 #' @rdname nlines
-setMethod("nlines", "NULL", function(x, colwidths, max_width, fontspec, col_gap = NULL) 0L)
+setMethod("nlines", "NULL", function(x, colwidths, max_width, fontspec = NULL, col_gap = NULL) 0L)
 
 #' @export
 #' @rdname nlines
-setMethod("nlines", "character", function(x, colwidths, max_width, fontspec, col_gap = NULL) {
-  if (missing(fontspec)) fontspec <- NULL
+setMethod("nlines", "character", function(x, colwidths, max_width, fontspec = NULL, col_gap = NULL) {
   splstr <- strsplit(x, "\n", fixed = TRUE)
   if (length(x) == 0) {
     return(0L)
--- a/R/generics.R
+++ b/R/generics.R
@@ -201,7 +201,7 @@ setGeneric(
 #' @rdname nlines
 setMethod(
   "nlines", "list",
-  function(x, colwidths, max_width, fontspec, col_gap = NULL) {
+  function(x, colwidths, max_width, fontspec = NULL, col_gap = NULL) {
     if (length(x) == 0) {
       0L
     } else {
@@ -223,7 +223,7 @@ setMethod("nlines", "character", function(x, colwidths, max_width, fontspec = NU
   }
 
   sum(vapply(splstr,
-    function(xi, max_width) {
+    function(xi, max_width, fontspec) {
       if (length(xi) == 0) {
         1L
       } else if (length(max_width) == 0) { ## this happens with strsplit("", "\n")
@@ -232,7 +232,8 @@ setMethod("nlines", "character", function(x, colwidths, max_width, fontspec = NU
         length(wrap_txt(xi, max_width, fontspec = fontspec))
       }
     }, 1L,
-    max_width = max_width
+    max_width = max_width,
+    fontspec = fontspec
   ))
 })
 
--- a/R/generics.R
+++ b/R/generics.R
@@ -202,6 +202,7 @@ setGeneric(
 setMethod(
   "nlines", "list",
   function(x, colwidths, max_width, fontspec, col_gap = NULL) {
+    if (missing(fontspec)) fontspec <- NULL
     if (length(x) == 0) {
       0L
     } else {
@@ -217,6 +218,7 @@ setMethod("nlines", "NULL", function(x, colwidths, max_width, fontspec, col_gap 
 #' @export
 #' @rdname nlines
 setMethod("nlines", "character", function(x, colwidths, max_width, fontspec, col_gap = NULL) {
+  if (missing(fontspec)) fontspec <- NULL
   splstr <- strsplit(x, "\n", fixed = TRUE)
   if (length(x) == 0) {
     return(0L)
--- a/R/tostring.R
+++ b/R/tostring.R
@@ -21,6 +21,7 @@ cwidth_inches_unsafe <- function(x) {
 #'
 #' @param fontspec (`font_spec`)\cr a font_spec object specifying the font information to use for
 #'   calculating string widths and heights, as returned by [font_spec()].
+#'   If `NULL`, no actions on the pdf device are taken and the function returns `FALSE`.
 #' @param silent (`logical(1)`)\cr If `FALSE`, the default, a warning will be
 #'   emitted if this function switches away from an active graphics device.
 #'
--- a/R/tostring.R
+++ b/R/tostring.R
@@ -21,7 +21,6 @@ cwidth_inches_unsafe <- function(x) {
 #'
 #' @param fontspec (`font_spec`)\cr a font_spec object specifying the font information to use for
 #'   calculating string widths and heights, as returned by [font_spec()].
-#'   If `NULL`, no actions on the pdf device are taken and the function returns `FALSE`.
 #' @param silent (`logical(1)`)\cr If `FALSE`, the default, a warning will be
 #'   emitted if this function switches away from an active graphics device.
 #'
--- a/R/tostring.R
+++ b/R/tostring.R
@@ -58,7 +58,7 @@ cwidth_inches_unsafe <- function(x) {
 #'
 #' @export
 open_font_dev <- function(fontspec, silent = FALSE) {
-  if (missing(fontspec) || is.null(fontspec)) {
+  if (is.null(fontspec)) {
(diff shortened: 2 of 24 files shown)


Melkiades (approved):
Great catch @averissimo! Amazing work. As far as I could see the error is coming from this function:
```r
ref_df_row <- function(row_path = NA_character_,
                       col_path = NA_character_,
                       row = NA_integer_,
                       col = NA_integer_,
                       symbol = NA_character_,
                       ref_index = NA_integer_,
                       msg = NA_character_,
                       max_width = NULL) {
  nlines <- nlines(msg, max_width = max_width)
```
@gmbecker could take a look at this, so we are sure that the above function does not need access to fontspec?

R/generics.R
pawelru, line 1: I think it would be good to document what `fontspec = NULL` means - it's not documented in the inherited parameter description:
```
#' @param fontspec (`font_spec`)\cr a font_spec object specifying the font information to use for
#'   calculating string widths and heights, as returned by [font_spec()].
```
Note that in the `open_font_dev()` (i.e. function from which we inherit `fontspec` param description), `fontspec` is required but this PR is all about making it optional. That makes me think that it won't be that easy to inherit

gmbecker: So there are a couple of things going on here:

Firstly, `fontspec` being `NULL` is valid; it essentially means "an arbitrary monospaced font of indeterminate size". In practice what this translates to is `nchar_ttype` (the thing that fontspec trickles down to that actually uses it)  falls back to `nchar`. This is done for speed in certain places, e.g., when we are calling `make_row_df` but we know we're not going to be doing anything with any of the columns that actually rely on `fontspec` (ie the extent and nlines ref type columns).

That said, I'm not sure this approach is the right one. it doesn't *really* make sense to ask how many lines a string will take up when printed without telling formatters what font you're going to print the said string in.

I'm much more inclined to think that the code that is managing to call nlines without a fontspec has a bug in it.

Note this should be invisible to end users, as the higher functions *they* are calling do have a default fontspec (its courier 8, not `NULL`), which should be passed down to everything that will eventually call nlines/nchar_ttype.

averissimo: > I think it would be good to document what fontspec = NULL means - it's not documented in the inherited parameter description:

Maybe we change the inherited param to add that `NULL` can be accepted and it will do nothing with fonts? From a superficial look at the functions where it's inherited a `NULL` value is valid (as @gmbecker also said)

It does affect a lot of function man pages :sweat_smile: 

image

> That said, I'm not sure this approach is the right one. it doesn't really make sense to ask how many lines a string will take up when printed without telling formatters what font you're going to print the said string in.

This was a naïve approach to fix the problem as it is breaking R CMD checks workflows on other packages.

**Feel free to take over this PR or with a new one with a better long-term solution for `{formatters}`**

note: an alternative (also naïve) fix could be to use `fontspec <- rlang::maybe_missing(fontspec, font_spec())` in `wrap_txt` to enforce `font_spec()` even when missing are passed through higher level functions.

This would add `{rlang}` as a dependency though.

shajoezhu (approved):
brilliant! Thanka @averissimo for the quick fix!

R/generics.R
Melkiades: @gmbecker let us know if you find a better solution to this issue. For me, the current fix works as it is probably happening only in that above mentioned case? I would go towards code freeze soonish

gmbecker:
@averissimo  @Melkiades  do you have reproducible cod ethat causes this? rtables gets through all its tests and vignettes for me without ever raising this error

averissimo:
@gmbecker, here:

``` r
formatters::nlines("01234567890", max_width = 5)
#> Error in FUN(X[[i]], ...): argument "fontspec" is missing, with no default
```

<sup>Created on 2024-06-10 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

averissimo:
The problem itself is not with a direct call like this one. But some packages' CI have started to fail on building vignettes or tests and I tracked the problem to this function.

[Here is a run](https://github.com/insightsengineering/teal.reporter/actions/runs/9447122586) where the problem was triggered by `{rlistings}` and the trace is available.

[Others runs](https://github.com/insightsengineering/tern.mmrm/actions/runs/9399866536/job/25888204566?pr=177) were failing from `{rtables}` (traceback also available)

gmbecker:
Again, I really think this is a version mismatch due to improper dependency caching in your automation

the current state of main on teal.reporter passes check for me locally:

```
gbecker@Gabriels-MacBook-Pro nestpkgs % R CMD check teal.reporter_0.3.1.9008.tar.gz
* using log directory ‘/Users/gbecker/gabe/checkedout/nestpkgs/teal.reporter.Rcheck’
* using R version 4.3.2 (2023-10-31)
* using platform: aarch64-apple-darwin20 (64-bit)
* R was compiled by
    Apple clang version 14.0.0 (clang-1400.0.29.202)
    GNU Fortran (GCC) 12.2.0
<~snip~>
* checking for unstated dependencies in examples ... OK
* checking installed files from ‘inst/doc’ ... OK
* checking files in ‘vignettes’ ... NOTE
Files named as vignettes but with no recognized vignette engine:
   ‘vignettes/_setup.Rmd’
(Is a VignetteBuilder field missing?)
* checking examples ... OK
* checking for unstated dependencies in ‘tests’ ... OK
* checking tests ...
  Running ‘testthat.R’
 OK
* checking for unstated dependencies in vignettes ... OK
* checking package vignettes in ‘inst/doc’ ... OK
* checking running R code from vignettes ...
  ‘previewerReporter.Rmd’ using ‘UTF-8’... OK
  ‘simpleReporter.Rmd’ using ‘UTF-8’... OK
  ‘teal-reporter-blocks-overview.Rmd’ using ‘UTF-8’... OK
  ‘teal-reporter.Rmd’ using ‘UTF-8’... OK
 NONE
* checking re-building of vignette outputs ... OK
* checking PDF version of manual ... OK
* DONE

Status: 1 NOTE
See
  ‘/Users/gbecker/gabe/checkedout/nestpkgs/teal.reporter.Rcheck/00check.log’
for details.


gbecker@Gabriels-MacBook-Pro nestpkgs % R -e "library(rtables); sessionInfo()" --vanilla
<~snip~>

> library(rtables); sessionInfo()
Loading required package: formatters
Loading required package: magrittr

Attaching package: ‘rtables’

The following object is masked from ‘package:utils’:

    str

R version 4.3.2 (2023-10-31)
Platform: aarch64-apple-darwin20 (64-bit)
Running under: macOS Sonoma 14.1

Matrix products: default
BLAS:   /Library/Frameworks/R.framework/Versions/4.3-arm64/Resources/lib/libRblas.0.dylib 
LAPACK: /Library/Frameworks/R.framework/Versions/4.3-arm64/Resources/lib/libRlapack.dylib;  LAPACK version 3.11.0

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

time zone: America/Los_Angeles
tzcode source: internal

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] rtables_0.6.7.9004    magrittr_2.0.3        formatters_0.5.7.9001

loaded via a namespace (and not attached):
 [1] compiler_4.3.2    backports_1.4.1   fastmap_1.1.1     cli_3.6.2        
 [5] tools_4.3.2       htmltools_0.5.8.1 grid_4.3.2        checkmate_2.3.1  
 [9] digest_0.6.34     lifecycle_1.0.4   rlang_1.1.3      
```

This is with the version of nlines that takes fonspec without a default value:

```
> library(rtables); formals(nlines)
Loading required package: formatters
Loading required package: magrittr

Attaching package: ‘rtables’

The following object is masked from ‘package:utils’:

    str

$x


$colwidths
NULL

$max_width
NULL

$fontspec


$col_gap
NULL

```

averissimo:
It was/is very hard for me to replicate the problem, even when using the CI images and running `staged.dependencies` in the container.

It's bizarre as it should pull the latest development versions. The latest run from `teal.reporter` installs the last commits from `main` in `{formatter}` and `{rtables}` via `{staged.dependencies}`

image

https://github.com/insightsengineering/teal.reporter/actions/runs/9462074037/job/26064237060

averissimo:
@gmbecker Thanks! you're right, it was a problem with versions (misconfiguration in `staged.dependencies` and a missed version bump in `DESCRIPTION`)

Is this fix still relevant? It appears to allow older versions of `rtables`/`rlistings` to play nicely with `{formatter}` (albeit not completely)

gmbecker:
> @gmbecker Thanks! you're right, it was a problem with versions (misconfiguration in `staged.dependencies` and a missed version bump in `DESCRIPTION`)
> 
> Is this fix still relevant? It appears to allow older versions of `rtables`/`rlistings` to play _(at least partially)_ nicely with `{formatter}`

I wouldn't want to bank on that with a lot more testing, I don't think its safe enough to say its supported. In other words it may happen to work but I'd consider anything that happens with version mismatches like this undefined behavior

PR #325, opened 2024-10-25
rm staged deps

### LINKED ISSUE #324: formatters

shajoezhu:
close #324


pawelru:
In the current form this would break CI. Unless you change `.github/workflows/check.yaml` file, staged.deps continue to be used and that file continue to be required.

shajoezhu:
> In the current form this would break CI. Unless you change `.github/workflows/check.yaml` file, staged.deps continue to be used and that file continue to be required.

sure. I will add the line 

      deps-installation-method: setup-r-dependencies

as I started with a few repos that were at the root of the dependency tree

pawelru:
This looks good! I can see that the new way is being used right now 👍

pawelru (approved):

PR #326, opened 2024-11-20
Fix bug in printing table with only one row, topleft information and column names with newlines

Melkiades:
Fixes https://github.com/insightsengineering/rtables/issues/942


edelarua (approved):
Looks great! Thanks @Melkiades

PR #333, opened 2025-01-14
fix: function crashes due to missing paginate argument

Tazovsky:
Function crashes, because `paginate` is not argument of function.

Error:

```
Error in isFALSE(paginate) : object 'paginate' not found
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: formatters
 Title: ASCII Formatting for Values and Tables
-Version: 0.5.10.9000
+Version: 0.5.10.9001
 Date: 2025-01-09
 Authors@R: c(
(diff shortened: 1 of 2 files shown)


shajoezhu:
hi @Tazovsky , can you share an example with us when you seeing the bug? Thanks!

shajoezhu:
i see! thanks for bring this to our attention. `export_as_rtf` is not a function that we actively testing, improvement is definitely needed. please can you share an example, I dont use rtf output, so we dont have any good example to reproduce, and debug this. thanks

shajoezhu:
hi @Tazovsky , can you update the doc please

shajoezhu:
close, and see https://github.com/insightsengineering/formatters/pull/343

Issue #336, opened 2025-01-31
support optional SAS/elementary school style rounding


gmbecker:
Support optional to-nearest-value (rather than the IEC  60559 to-nearest-even) rounding.

I will prepare a PR.

Default behavior will not change so this will be fully backwards-compatible.

shajoezhu:
closed by #337

Issue #319, opened 2024-09-10
Feature: when formatting function has na_str as function argument, allow format_value to pass this argument


iaugusty:
**Summary**
We'd like to use NE as our default na_str character, in combination with a formatting function, rather than the built-in character string formats.

On the other hand, we want to maintain consistent approach in switching to another na_str character in rtables layouts, eg pass the parameter na_str into an analyze call.

eg
```r
custom_fmt_function <- function(xxx, .na_str = "NE") {
 mainfun <- function(x, output, na_str = .na_str){}
mainfun
}

analyze(vars = "VAR",
.stats = "mystat",
.formats = c( "mystat" = custom_fmt_function(xxx)),
na_str = "ne")

rather than 

analyze(vars = "VAR",
.stats = "mystat",
.formats = c( "mystat" = custom_fmt_function(xxx, .na_str = "ne"))
```

Melkiades:
I think this is a relevant update and we should unify our behavior in someway

gmbecker:
I will be preparing  a PR that implements optional SAS/elementary school style rounding, and this can be easily rolled into that because it will involve passing around a round_type argument which we also want the format_function to be able to optionally accept. @Melkiades  @shajoezhu

shajoezhu:
closed by https://github.com/insightsengineering/formatters/pull/337

PR #337, opened 2025-01-31
implement optional 'sas' rounding, pass round_type and na_str to fmt fun

gmbecker:
addresses #336 and #319


gmbecker:
@cicdguy looks like the commit pushed by the restyle files job is failing the style check

gmbecker:
Hmm, it'd be nice if we could get rid of the x. the PR is still showing as "not passing". @shajoezhu @Melkiades sounds like this PR is "effectively passed" and ready for review.

shajoezhu:
> Hmm, it'd be nice if we could get rid of the x. the PR is still showing as "not passing". @shajoezhu @Melkiades sounds like this PR is "effectively passed" and ready for review.

you could do a "empty commit" to trigger the pipeline after the CI restyle btw.

shajoezhu:
Hi @gmbecker , I would suggest in https://github.com/insightsengineering/scda.test/pull/183, if you add a snapshot test, I know that jnj don't have this ready yet, so I think it would be nice if you could add a snapshot test into scda.test, so in the future, we are aware that new changes will cause breaking change unnoticed.

Melkiades:
@gmbecker could you please update the news file?

PR #342, opened 2025-02-10
add nearest-value rounding support

Melkiades:
Alternative work for #336. This PR does not need any change downstream.

Please, @gmbecker, let me know what you think!

@shajoezhu, @edelarua, and @ayogasekaram fyi


gmbecker:
@Melkiades I am extremely wary of having a global option that controls rounding behavior, because it damages reproducibility.

Imagine I get an edge-case number (one that rounds to a different value under sas and iec rounding) and I send my data and script to someone else, but the option was set to sas rounding somewhere else, e.g. in another output's script. When they run the script on the data they will get an output that performs iec rounding when rendered, and so won't agree with what I got.

I'd much prefer that formatters not provide an option like that. chevron could (although it doesn't need to because the default behavior remains the iec rounding that roche uses) and our J&J exporters will all be setting round_type to sas, but I still think it should be an argument that is passed down and its up to the user or calling code to set it correctly.

shajoezhu:
Hi @gmbecker , I also would prefer @Melkiades 's implementation, as I suggested in the email last year. I think using an option is the most efficient way for now. 

@Melkiades 's change only appeared in three files, and very limited function, whereas the proposal in #337 has a lot of downstream changes in functions (5 files), and will have to change rtables as well (https://github.com/insightsengineering/rtables/pull/990/files), we have several packages (internally as well, use toString and matrix_form functions), the change you had proposed is going to have a lot of downstream impact on us for changing, as well as maintaining. 

We have very established pipeline on this, and I am afraid the risk is very high for us to accept your PR as it is.

Would it be helpful for jnj to consider using a chevron like approach to control these options, as I understand it is still at an exploring stage for jnj, I feel you guys have more flexibilities, and open minded with these proposals. Thanks! @khatril  @iaugusty fyi

gmbecker:
@shajoezhu @Melkiades let me try again to explain why I think this approach is a mistake.

This software stack is used to generate outputs submitted to the FDA and other regulatory authorities. Code is also submitted, AFAIK, so that FDA personnel can check the outputs. If the outputs the fda checkers get different numbers when they run the code, that would be, to put it mildly, "not good".  Having a global option (which, I should note, there's no meaningful way to require it be set in any particular script), opens the possibility of that happening.

> @Melkiades 's change only appeared in three files, and very limited function, whereas the proposal in #337 has a lot of downstream changes in functions (5 files), and will have to change rtables as well (https://github.com/insightsengineering/rtables/pull/990/files), we have several packages (internally as well, use toString and matrix_form functions), the change you had proposed is going to have a lot of downstream impact on us for changing, as well as maintaining.

I prepared the relevant downstream PRs, and the change is fully backwards compatible, so you guys won't need to do anything if you don't want to support sas-style rounding in your downstream uses.

> 
> We have very established pipeline on this, and I am afraid the risk is very high for us to accept your PR as it is.

Maybe we should discuss this in a meeting, because I don't see the risk, the default value for round_type is the backwards compatible iec rounding so it shouldn't break any code (and doesn't cause failures in scda.test for me).

PR #356, opened 2025-07-17
Fix pagination listings with newlines

Melkiades:
https://github.com/insightsengineering/rlistings/issues/269


Melkiades:
@shajoezhu @edelarua @ayogasekaram Fix should be ready now! Tested it on scda.test and it Passes

edelarua (approved):
Lgtm!!

PR #377, opened 2026-09-21
Validation report

shajoezhu:
**What changes are proposed in this pull request?**

* Adds a "Validation report" entry to the pkgdown navbar, resolved per documentation version. (#377)

pkgdown builds one static site per version, so a hardcoded link would freeze at whichever report was current when it was written. `pkgdown/extra.js` reads the version from the navbar and points the link at the newest report at or below it. Versions older than the first report drop the item entirely. `_pkgdown.yml` keeps a static href as the fallback for readers with JavaScript disabled.

Reports currently exist for 0.5.11 and 0.5.12, so a 0.5.13 site links to the 0.5.12 report.

**Also in this PR:** `DESCRIPTION` moves to `Version: 0.5.13` / `Date: 2026-09-15` and the `NEWS.md` heading follows. `v0.5.13` was tagged and released on 2026-09-15 but `main` was left at `0.5.12.9003`, so this brings the branch back in line with the published release. Flagging it because it is not implied by the PR title.


Melkiades:
@shajoezhu I pushed 8b89f3c9 to this branch and filled in the description, since it was empty and the PR also carries the 0.5.13 version sync. Shout if you would rather I had left either alone.

Three small things in `extra.js`:

* `findLink()` scanned every anchor on the page, so a vignette linking the words "validation report" would have been rewritten instead of the navbar item. Scoped it to the navbar.
* `closest()` and `remove()` were the only ES6-era DOM calls in an otherwise strict ES5 file. Swapped for equivalents so the target is consistent.
* `pickReport()` took the last match, so it silently depended on `REPORTS` being sorted ascending. It now takes the maximum instead. I also corrected the header comment: adding a version to `REPORTS` does need one other change, the static href in `_pkgdown.yml` that no-JS readers get.

Checked the logic with `REPORTS` both sorted and deliberately reversed, same results throughout: `0.5.10` no link, `0.5.11` and `0.5.12` their own, `0.5.13` and `0.6.0` to `0.5.12`. Also confirmed against the live nesttemplate navbar that `.navbar small` really does hold the version, and that the 0.5.13 report is a 404 so stopping `REPORTS` at 0.5.12 is right.

Nice approach, resolving per version is the right call given pkgdown builds a site per tag.

One I left alone for you to decide: `0.5.11-rc1` resolves to the 0.5.11 report, because `parseInt("rc1")` is `NaN` and becomes 0, so the rc compares equal to the release. The rc doc sites are real, so it is reachable. Pointing an rc at its eventual release report seemed more useful than no link, but happy to change it.

Melkiades (approved):
lgtm! thanks Joe

